PR #3, opened 2021-10-04
more informative README

rkurchin:
Thoughts?


rkurchin:
The badge links are also broken...I replace `mfherbst` with `JuliaMolSim` but obviously there's no actual docs deployment etc. yet so perhaps we should just remove those for now?

jgreener64:
Looks good to me.

jgreener64 (approved):

Gregstrq:
I have a few comments.
- If we want to support struct-of-arrays and array-of-structs on equal footing, we need to emphasize the dispatches of the functions on the system:
```
get_index(::AbstractSystem, index)
get_position(::AbstractSystem, index)
get_element(::AbstractSystem, index)
get_velocity(::AbstractSystem, index)
```
In this regards, the dispatches on `AbstractParticle` are the ones that are actually optional.

- Do we really need a special `AbstractParticle` type? The way README is written, It looks like we need to implement two concrete subtypes of `AbstractParticle`: one which contains coordinates and Element, and one which contains coordinates, Element and velocities. These two subtypes partially overlap in functionality which looks ugly. Why can't `get_index` simply return a tuple. 
In the first case it would be `(element, coordinate)`, in the second case `(element, coordinate, velocity)`.

README.md
```diff
@@ -1,6 +1,46 @@
...
+
+## Overview
+AtomsBase defines a few abstract types used for specifying an atomic system. We will describe them briefly here, from the "top down." Users and/or prospective developers may also find `implementation_simple.jl` a useful reference for a simple concrete implementation of the interface.
+### System
+An object describing a system should be a subtype of `AbstractSystem` and will in general store identifiers, positions, and (if relevant) velocities of the particles that make it up. It takes a type parameter (which must be `<:AbstractParticle`, see below) to indicate what types of particles these are, and requires dispatch of the following functions:
+* `get_box(::AbstractSystem)::Vector{<:AbstractVector}`: should return a set of basis vectors describing the coordinate system of the particles
+* `get_boundary_conditions(::AbstractSystem)::AbstractVector{BoundaryCondition}`: returns the boundary conditions corresponding to each spatial dimension of the system (see below for more on the `BoundaryCondition` type)
```
mfherbst, line 23: maybe emphasise the previous point: at the boundary of the box.

mfherbst:
> If we want to support struct-of-arrays and array-of-structs on equal footing, we need to emphasize the dispatches of the functions on the system:

I agree we should emphasise that part (also with a concrete implementation ideally).

> In the first case it would be (element, coordinate), in the second case (element, coordinate, velocity).

These are two different interfaces, so that needs to be reflected by the (abstract) type of the first argument to `getindex`. Also it makes it very cumbersome to introduce another field on the same footing like a coordinate or velocity in a user-specified field (e.g. like an "acceleration" (didn't come up with something more clever)). Because then you would need to return a 4-tuple instead of a 3-tuple.

> one which contains coordinates and Element, and one which contains coordinates, Element and velocities. 

I'm not sure I get your point @Gregstrq, but if you don't need / want velocities, you don't need to store them (and just return `missing` as appropriate). I understand that returning an array of missing in `get_velocity(::AbstractSystem, index)` is not super great, though.

README.md
```diff
@@ -17,19 +17,30 @@ Currently, the design philosophy is to be as lightweight as possible, with only
...
+
 ### Particles
 Particle objects are subtypes of `AbstractParticle`, and also take a type parameter that is `<:AbstractElement` (see below) to indicate how particles are identified.
 
-The interface is flexible to an `AbstractSystem` subtype being a "struct-of-arrays" or "array-of-structs" implementation. In the former case, particle objects would only ever be explicitly constructed when `getindex` is invoked on the system, as a "view" into the system.
+In the SoA case, particle objects would only ever be explicitly constructed when `getindex` is invoked on the system, as a "view" into the system.
```
Gregstrq, line 43: If this is true then the appearance of `AbstractParticle` as a type parameter of `AbstractSystem` looks strange in the SoA case

src/interface.jl
```diff
@@ -90,30 +90,30 @@ Base.setindex!(::AbstractSystem, ::Int) = error("AbstractSystem objects are not
...
+    print(io, "Particle(", element(part), ") @ ", position(part))
 end
 function Base.show(io::IO, ::MIME"text/plain", part::AbstractAtom)
-    print(io, "Atom(", get_atomic_symbol(part), ") @ ", get_position(part))
+    print(io, "Atom(", atomic_symbol(part), ") @ ", position(part))
 end
 function Base.show(io::IO, mime::MIME"text/plain", sys::AbstractSystem)
     println(io, "System:")
-    println(io, "    BCs:        ", get_boundary_conditions(sys))
-    println(io, "    Box:        ", get_box(sys))
```
mfherbst, line 116: box -> bounding_box ?

README.md
```diff
@@ -1,6 +1,57 @@
...
+
+A way to think about this broadly is that the difference amounts to the ordering of function calls. For example, to get the position of a single particle in an AoS implementation, the explicit funciton chaining would be `position(getindex(sys))` (i.e. extract the single struct representing the particle of interest and query its position), while for SoA, one should prefer `getindex(position(sys))` (extract the array of positions, then index into it for a single particle). The beauty of an abstract interface in Julia is that these details can be, in large part, abstracted away through method dispatch such that the end user sees the same expected behavior irrespective of how things are implemented "under the hood."
+
+### System
+An object describing a system should be a subtype of `AbstractSystem` and will in general store identifiers, positions, and (if relevant) velocities of the particles that make it up, as well as a set of coordinate bounds.
+
+An `AbstractSystem` takes a type parameter (which must be `<:AbstractParticle`, see below) to indicate what types of particles these are, and requires dispatch of the following functions:
```
Gregstrq, line 28: We need to describe the type parameters of `AbstractSystem` since we have merged it.

README.md
```diff
@@ -1,6 +1,57 @@
...
+* `box(::AbstractSystem)::Vector{<:AbstractVector}`: should return a set of basis vectors describing the boundaries of the coordinates in which the system resides
```
Gregstrq, line 29: May be change it to
```
box(::AbstractSystem{D})::SVector{D, SVector{D, <:Unitful.Length}}
```

README.md
```diff
@@ -1,6 +1,57 @@
...
+* `boundary_conditions(::AbstractSystem)::AbstractVector{BoundaryCondition}`: returns the boundary conditions corresponding to each spatial dimension of the system (see below for more on the `BoundaryCondition` type)
```
Gregstrq, line 30: Again, update it to reflect the current state:
```
boundary_conditions(::AbstractSystem{D})::SVector{D, BoundaryCondition}
```

README.md
```diff
@@ -1,6 +1,57 @@
...
+
+`AbstractSystem` subtypes `AbstractVector`, thus the following functions must also be dispatched:
```
Gregstrq, line 32: Since we no longer subtype `AbstractVector`, we need to change this line

README.md
```diff
@@ -1,6 +1,57 @@
...
+* `Base.getindex(::AbstractSystem, ::Int)`
+* `Base.size(::AbstractSystem)`
+
+By default, the `position` and `velocity` functions dispatch onto `AbstractSystem` objects as a broadcast over the system (e.g. `position(sys::AbstractSystem) = position.(sys)`, invoking the dispatch of `position` onto the `AbstractParticle` type parameter of the system). In an SoA implementation, custom dispatches should probably be included to avoid the construction of the particle objects.
```
Gregstrq, line 36: Well, I think `position` and `velocity` here need to be renamed to `positions` and `velocities`.
If we remove `get_` from `get_position` and `get_velocity`, we get a huge problem.

PR #8, opened 2021-10-12
specific getindex for single vs. slice

rkurchin:
Thought it might make sense to explicitly include both of these options, per our discussion last Friday. I think we should also split out the sample implementations and rename the current `implementation_simple` to `implementation_simple_aos` or something like that and then add an analogous SoA version, which would need to explicitly dispatch these separately.

Not sure this actually makes sense to merge as-is though, since currently I think it would also break the existing `implementation_simple`, and we may want to have things set up for that case to be able to "just work." If so, I think we'd just need to define the slice dispatch to be the standard Julia behavior.

So yeah, this is more to start a conversation I suppose? But getting it to a non-draft state should be pretty simple once we decide what we want the behavior to be.


rkurchin:
Yeah, that was sort of what I was thinking as well... 🤔

rkurchin:
Linear indexing is addressed in #23, closing this for now and opening an issue to potentially support Cartesian indexing in the future.

PR #10, opened 2021-10-12
`AbstractSystem` doesn't subtype anything but supports indexing interface

### LINKED ISSUE #6: `AbstractSystem` as a subtype of `AbstractArray`
Right now, `AbstractSystem` is defined as a subtype of `AbstractVector`. In case of a static `D`-dimensional lattice it is more convenient to use cartesian indices instead of linear indices, so this subtyping does not make much sense in this particular case.

Instead, `AbstractSystem` could subtype `AbstractArray` ( or not subtype anything).
In either case, we keep the functions `Base.getindex` and `Base.size`. I also propose to add `Base.length` to the list of interface functions because `Base.size` and `Base.length` have different meaning in the context of multidimensional arrays.

rkurchin:
Resolves #6 

Is this basically what you were thinking, @Gregstrq? (The second commit is kind of take-it-or-leave-it IMO and also I didn't test it yet haha, just put it there to point out that we could do that and it might be useful...)

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -1,6 +1,7 @@
 using Unitful
 using UnitfulAtomic
 using PeriodicTable
+using StaticArrays
 
 export AbstractElement, AbstractParticle, AbstractAtom, AbstractSystem
 export Element
@@ -73,9 +74,9 @@ struct Periodic      <: BoundaryCondition end  # Periodic BCs
 # The system type
 #     Again readonly.
 #
-abstract type AbstractSystem{AT <: AbstractParticle,D} <: AbstractArray{AT,D} end
-get_box(::AbstractSystem)::Vector{<:AbstractVector} = error("Implement me")
-get_boundary_conditions(::AbstractSystem)::AbstractVector{BoundaryCondition} = error("Implement me")
+abstract type AbstractSystem{AT <: AbstractParticle, D} <: AbstractArray{AT,D} end
+get_box(::AbstractSystem)::SVector{D, SVector{D, Unitful.Length}} = error("Implement me")
+get_boundary_conditions(::AbstractSystem)::SVector{D,BoundaryCondition} = error("Implement me")
 get_periodic(sys::AbstractSystem) = [isa(bc, Periodic) for bc in get_boundary_conditions(sys)]
 
 # Note: Can't use ndims, because that is ndims(sys) == 1 (because of AbstractVector interface)
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -80,7 +80,7 @@ get_boundary_conditions(::AbstractSystem)::SVector{D,BoundaryCondition} = error(
 get_periodic(sys::AbstractSystem) = [isa(bc, Periodic) for bc in get_boundary_conditions(sys)]
 
 # Note: Can't use ndims, because that is ndims(sys) == 1 (because of AbstractVector interface)
-n_dimensions(::AbstractSystem{T,N}) where {T,N} = N
+n_dimensions(::AbstractSystem{T,D}) where {T,D} = D
 
 Base.getindex(::AbstractSystem, ::Int)  = error("Implement me")
 Base.size(::AbstractSystem)             = error("Implement me")
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -74,7 +74,7 @@ struct Periodic      <: BoundaryCondition end  # Periodic BCs
 # The system type
 #     Again readonly.
 #
-abstract type AbstractSystem{AT <: AbstractParticle, D} <: AbstractArray{AT,D} end
+abstract type AbstractSystem{AT <: AbstractParticle, D} end
 get_box(::AbstractSystem)::SVector{D, SVector{D, Unitful.Length}} = error("Implement me")
 get_boundary_conditions(::AbstractSystem)::SVector{D,BoundaryCondition} = error("Implement me")
 get_periodic(sys::AbstractSystem) = [isa(bc, Periodic) for bc in get_boundary_conditions(sys)]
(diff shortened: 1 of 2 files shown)


src/interface.jl
```diff
@@ -73,16 +74,17 @@ struct Periodic      <: BoundaryCondition end  # Periodic BCs
 # The system type
 #     Again readonly.
 #
-abstract type AbstractSystem{AT <: AbstractParticle} <: AbstractVector{AT} end
-get_box(::AbstractSystem)::Vector{<:AbstractVector} = error("Implement me")
-get_boundary_conditions(::AbstractSystem)::AbstractVector{BoundaryCondition} = error("Implement me")
+abstract type AbstractSystem{AT <: AbstractParticle, D} <: AbstractArray{AT,D} end
```
Gregstrq, line 77: At the end of the previous meeting, we discussed that it might be a good idea not to subtype `AbstractArray` at all.
For example, for molecular dynamic simulations or for a static molecule there is no point to index into system using Cartesian indices. Therefore, matching dimensionality of `AbstractSystem` to dimensionality of `AbstractArray` is a bad thing. Then, if we do not subtype `AbstractArray`, we do not actually need to interchange the order of `D` and `AT`.

The second problem is that using `AbstractParticle` as type parameter skews the balance in favour of array of structs.

So, I propose we write this line as
```
abstract type AbstractSystem{D, ET<:AbstractElement} end
```

src/interface.jl
```diff
@@ -73,16 +74,17 @@ struct Periodic      <: BoundaryCondition end  # Periodic BCs
...
+get_box(::AbstractSystem)::SVector{D, SVector{D, Unitful.Length}} = error("Implement me")
```
Gregstrq, line 78: May be it is a good idea to combine several `SVector`-s into single `SMatrix`?
```
get_box(AbstractSystem{D})::SMatrix{D,D,<:Unitful.Length} where {D} = error("Implement me")
```

src/interface.jl
```diff
@@ -73,16 +74,17 @@ struct Periodic      <: BoundaryCondition end  # Periodic BCs
...
+get_boundary_conditions(::AbstractSystem)::SVector{D,BoundaryCondition} = error("Implement me")
 get_periodic(sys::AbstractSystem) = [isa(bc, Periodic) for bc in get_boundary_conditions(sys)]
 
 # Note: Can't use ndims, because that is ndims(sys) == 1 (because of AbstractVector interface)
-n_dimensions(sys::AbstractSystem) = length(get_boundary_conditions(sys))
+n_dimensions(::AbstractSystem{T,N}) where {T,N} = N
```
Gregstrq, line 83: We should keep the type parameters consistent: either use everywhere `D` or `N` for dimensionality.
Since you used `D` in line 77, let us stick with it. Considering we do not subtype `AbstractArray` and do not reorder type parameters, we can write
```
```

src/interface.jl
```diff
@@ -73,16 +74,17 @@ struct Periodic      <: BoundaryCondition end  # Periodic BCs
 # The system type
 #     Again readonly.
 #
-abstract type AbstractSystem{AT <: AbstractParticle} <: AbstractVector{AT} end
-get_box(::AbstractSystem)::Vector{<:AbstractVector} = error("Implement me")
-get_boundary_conditions(::AbstractSystem)::AbstractVector{BoundaryCondition} = error("Implement me")
+abstract type AbstractSystem{AT <: AbstractParticle, D} <: AbstractArray{AT,D} end
+get_box(::AbstractSystem)::SVector{D, SVector{D, Unitful.Length}} = error("Implement me")
+get_boundary_conditions(::AbstractSystem)::SVector{D,BoundaryCondition} = error("Implement me")
```
Gregstrq, line 79: Since boundary conditions can be different subtypes of `BoundaryCondition` we can't really dispatch on `SVector{D, BoundaryCondition}`. `Tuple` looks more suitable in this case:
```
get_boundary_conditions(::AbstractSystem{D})::NTuple{D, BoundaryCondition} = error("Implement me")
```

src/interface.jl
rkurchin: I agree with the second point, but I personally think there is still utility to subtyping `AbstractArray`. As was discussed on #6, for non-Cartesian indices, the array can just be of one dimension, but having a "canonical ordering" to the particles (even if arbitrary) is still useful.

rkurchin: Yeah that's true I suppose, we could support `getindex` without actually subtyping...I guess I'd be fine with that. Curious what others things, cc @jgreener64 @mfherbst

src/interface.jl
rkurchin: yeah if we do want to do this with static sizes, that would probably make sense (I was just templating it off of what @mfherbst had already done in `implementation_simple`). Though the "risk" there is you lose the clarity of index ordering...I personally often get confused with these matrices if the cell vectors are the columns or the rows 😆

rkurchin:
I guess one "open" question still is whether we want this to be with StaticArrays or whether I should revert that commit.

Gregstrq:
I think the most crucial question is about `AT<:AbstractParticle` type parameter.

>I guess one "open" question still is whether we want this to be with StaticArrays or whether I should revert that commit.

If you are referring to the boundary conditions, then my suggestion is not to go back to `AbstractVector{BoundaryCondition}`, but replace `SVector{D, BoundaryCondition}` with ​`NTuple{D, BoundaryCondition}`.

>and, relatedly, the matrix vs. vector of vectors question for the box

Yeah, it would be nice to get to an agreement regarding this too.

rkurchin:
@Gregstrq, I think the type parameter can be a separate discussion (there's already an issue for it). This PR doesn't change that behavior at the moment, it just removes subtyping of `AbstractVector`. Similarly, I think the boundary conditions are also a separate discussion from this.

The matrix vs. vector thing does fit here IMO, so I'm curious for others' thoughts on that, e.g. @mfherbst since he put it that way in the implementation sample.

src/interface.jl
mfherbst: I can give you none out of the top of my head, but it turns out people disagree between arranging vectors as rows and columns ... and this frequently causes trouble. So by basing the interface as vectors of vectors this aspect is unambiguous.

PR #13, opened 2021-10-15
Minor technical improvements

Gregstrq:
- removed unnecessary type parameters in functions that dispatch on `AbstractSystem`
- made `AbstractAtom` an alias
- Fixed non-concrete coordinates in `SimpleAtom` and `SimpleSystem`


src/interface.jl
```diff
@@ -37,7 +37,7 @@ get_atomic_mass(el::ChemicalElement)   = el.data.atomic_mass
 abstract type AbstractParticle{ET<:AbstractElement} end
 get_velocity(::AbstractParticle)::AbstractVector{<: Unitful.Velocity} = missing
 get_position(::AbstractParticle)::AbstractVector{<: Unitful.Length}   = error("Implement me")
-(get_element(::AbstractParticle{ET})::ET) where {ET<:AbstractElement} = error("Implement me")
+get_element(::AbstractParticle{ET})::ET where {ET<:AbstractElement} = error("Implement me")
```
rkurchin, line 40: those parentheses are actually necessary...otherwise when you try to import the package you get a `ET not defined` error.

Issue #1, opened 2021-10-02
Restrictions for coordinate types


Gregstrq:
This is the place to discuss whether we want to enforce specific types for the coordinates and, by extension, for the velocities.

Right now, the proposed way is to assume that the coordinates must have a unitful type with dimension of the length, while the velocities must have a unitful type with the corresponding dimension of the velocity. As a consequence, all of the realizations of the abstract interface need to enforce this assumption.

What is the opinion of the community members about this restriction?

My take on it is that this is an unnecessary restriction. Although I agree that the majority of use cases pertain to the particles in real space, I can imagine doing dynamical simulations with the particles living in some abstract space. For example, in momentum space. As a result, I propose to relax the restriction and allow the coordinates and velocities to have arbitrary types. Instead, I propose that we should enforce the "homogeneity" of the used types: one should not be able to mix the unitless and unitful coordinates while describing the same system (one should not be able to use unitful quantities with different dimensions as well).

eahenle:
In the case of crystalline systems, coordinates are often fractional, so I think allowing unitless is a good thing.

rkurchin:
This is absolutely true, but our thought was that even in the case of fractional coordinates, there is a set of basis vectors to which those fractions refer, and those vectors have units. For the purposes of interoperability that we hope to facilitate by this package (e.g. passing between MD/DFT/ML, or doing file I/O, etc.), I think it makes sense that this function would return unitful Cartesian coordinates in that case.

(Of course, an individual implementation of the interface could have other functions for passing around fractional coordinates internally, and would almost certainly need to have them!)

Does this make sense?

jgreener64:
Initially I thought no units for the reasons given above. Now I have switched to the other view, for the reasons Rachel outlines.

We should keep in mind what an interface is intended to do: to provide functions with guarantees to downstream users. It doesn't stop people writing their own functions to return positions in momentum space or fractional coordinates. It just says that if you want to implement the interface, you need to have some function that converts to Cartesian space to help downstream users. There are a limited number of cases where this may be impossible (e.g. particles in momentum space), but the interface can't support everything and I feel Cartesian coordinates are important enough to be a first class citizen.

A couple of use cases where relaxing unit requirements would make the interface less powerful:
- Writing a file format that uses Angstrom as length units. With a length dimension guaranteed in the position you can always convert correctly, without it you are writing values blind.
[...]

PR #16, opened 2021-10-16
Don't enforce Unitful coordinates

rashidrafeek:
With regard to #1, I am just putting this here as a draft to discuss why this might be better than enforcing Uniful coordinates.

### DIFF
--- a/src/implementation_simple.jl
+++ b/src/implementation_simple.jl
@@ -4,7 +4,7 @@ using StaticArrays
 
 export SimpleAtom, SimpleSystem, SimpleAtomicSystem
 
-struct SimpleAtom{D, T<:Unitful.Length} <: AbstractAtom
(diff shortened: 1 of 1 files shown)


Gregstrq:
If we decide to relax Unitful coordinates, we also need to do it for the return types in the interface.

Anyway, since I was the one pushing for not enforcing this, I am a bit biased. So we need to see what @rkurchin, @mfherbst and @jgreener64 have to say.

rashidrafeek:
I have replied in that issue itself (https://github.com/JuliaMolSim/AtomsBase.jl/issues/1#issuecomment-944877602) on why I think this might be better. Maybe it might be better to have the discussion there. And we could discuss on the specifics of implementation after we reach a decision in #1.

cortner:
I see there there was a PR once upon a time. Too late to re-open, but this is exactly the change that is needed.

PR #21, opened 2021-11-02
GitHub Actions for autobuild, TagBot and CompatHelper

jgreener64:
This PR adds a few package related things:
- A GitHub Action to run the package tests on pushing to branches, and every week as a cron job. Our tests are currently minimal but even checking the package can load is something, and we may add more tests later.
- TagBot GitHub Action to make tagging releases easier.
- CompatHelper GitHub Action to check for updates in our dependencies. Compat bounds are added to Project.toml to allow this (which also makes package registration faster).

The registrator bot is also now active on the repo.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,49 @@
+name: CI
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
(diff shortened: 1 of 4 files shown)


rkurchin:
This seems sufficiently uncontroversial that I'm just going to merge it.

PR #23, opened 2021-11-10
simplify types

rkurchin:
Starting to work on this. Will try to commit-by-commit remove more things (and only ever committing a working state) so if we decide we want to go only part of the way, it's easy to just revert. This present commit removes only the `AbstractElement` type, which, now that I've done it, seems like unambiguously a good idea and I don't think it was adding much of anything.


rkurchin:
Okay, there we have it! Personally, I feel quite good things up through fede982, but I'm on the fence about also removing `AbstractSystem`...I can't decide if what we gain in flexibility offsets what we would lose in being able to dispatch things with some guarantees about things like dimensionality, what sorts of particles make up the system, etc., the way we can with an `AbstractSystem` that would have those things as type parameters. Eager to discuss.

I've also made some notes to myself about other potential small tweaks, the applicability of some of which will depend on how many commits in this we keep, so I'll do them after we decide rather than cluttering up the history here.

jgreener64 (approved):
Looks good to me, readme and tests are nice too.

README.md
```diff
@@ -12,46 +12,49 @@ AtomsBase is an abstract interface for representation of atomic geometries in Ju
...
+### System state and properties
+The only required properties to be specified of the system is the species of each component of the system and the positions and velocities associated with each component. These are accessed through the functions `species`, `position`, and `velocity`, respectively. The default dispatch of these functions onto an `AbstractSystem` object is as a broadcast over it, which will "just work" provided the indexing/iteration interfaces have been implemented (see above) and the functions are defined on individual system components.
```
Gregstrq, line 37: Functions `species`, `position` and `velocity` are also supposed to have indexing version.
Please, emphasize it in Readme.

Gregstrq (changes requested):
We need to add indexing versions of `position` and `species` for the concrete implementations.

README.md
rkurchin: they're in another PR, #8, which still has some kinks to be worked out IIRC

Gregstrq: We should also point out in line 55 of Readme that two proposed chainings are different implementations of the indexing versions of `something` functions.

rkurchin: If we want to just decide that we're only going to support linear indexing for now, though, I can add that in and maybe just close the other one and open an issue instead

rkurchin: Okay, so do we want to just do linear indexing for now?

mfherbst: I think just linear indexing for now is fine.

mfherbst (commented):
A few nits, but generally LGTM.

src/interface.jl
```diff
@@ -81,36 +46,44 @@ struct DirichletZero <: BoundaryCondition end  # Dirichlet zero boundary (i.e. m
...
-# Note: Can't use ndims, because that is ndims(sys) == 1 (because of AbstractVector interface)
+# Note: Can't use ndims, because that is ndims(sys) == 1 (because of indexing interface)
 n_dimensions(::AbstractSystem{D}) where {D} = D
 
-
-# indexing interface
+# indexing and iteration interface
 Base.getindex(::AbstractSystem, ::Int) = error("Implement me")
-Base.size(::AbstractSystem) = error("Implement me")
 Base.length(::AbstractSystem) = error("Implement me")
```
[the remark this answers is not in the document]
rkurchin: I can see it both ways. On the one hand, it's redundant to throw an error "manually" here, but on the other, having all the required functions in the `interface.jl` file does make it a useful reference for someone who wants to sit down and implement the interface...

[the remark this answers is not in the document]
jgreener64: I would lean towards not putting the errors here, and perhaps adding a comment if we want the information to be in that file. See the first section of https://www.oxinabox.net/2020/04/19/Julia-Antipatterns.html for some discussion.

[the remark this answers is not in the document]
mfherbst: I agree with @jgreener64 (and oxinabox) on that point very much.

mfherbst (approved):
lgtm

src/implementation_aos.jl
```diff
@@ -1,32 +1,48 @@
...
+
 using StaticArrays
 
 export FlexibleSystem
 
 # TODO Switch order of type arguments?
-struct FlexibleSystem{D,ET<:AbstractElement,AT<:AbstractParticle{ET}} <: AbstractSystem{D,ET,AT}
-    box::SVector{D,<:SVector{D,<:Unitful.Length}}
+struct FlexibleSystem{D,S,L<:Unitful.Length,AT} <: AbstractSystem{D,S}
+    box::SVector{D,<:SVector{D,L}}
```
cortner, line 11: "morally" it seems to me that `box` and `boundary_conditions` should be tuples of ... rather than SVectors. But I don't have too strong a view on that.

src/implementation_aos.jl
```diff
@@ -1,32 +1,48 @@
...
+    }(
+        box,
+        boundary_conditions,
+        particles,
+    )
 end
 # convenience constructor where we don't have to preconstruct all the static stuff...
 function FlexibleSystem(
-    box::Vector{Vector{L}},
+    box::Vector{<:AbstractVector{L}},
```
cortner, line 27: why type this? the box could be provided as a Vector, AbstractVector, Tuple, .... same with bc and particles below. This is a convenience constructor no? I would then just make it convenient to construct the thing and enforce such strict typing.

src/atoms.jl
```diff
@@ -0,0 +1,35 @@
...
+
+function StaticAtom(position, symbol::Union{Integer,AbstractString,Symbol,AbstractVector})
+    StaticAtom(position, elements[symbol])
+end
+
+function Base.show(io::IO, a::StaticAtom)
+    print(io, "StaticAtom: $(a.element.symbol)")
+end
+
+const AbstractAtomicSystem{D} = AbstractSystem{D,Element}
```
cortner, line 24: I don't have time to look for the definition of AbstractSystem, so I'll add this comment here: I can sort of see why one might want to make the dimension a type parameter, though I'm not entirely sure where this is needed? 

But does the element type really have to be a type parameter in the abstract type? What is the use case?

src/implementation_soa.jl
```diff
@@ -1,18 +1,19 @@
-# Example implementation using as few function definitions as possible
 #
+# Implementation of AtomsBase interface in a struct-of-arrays style.
+#
+
 using StaticArrays
 
 export FastSystem
```
cortner, line 7: I find the `FlexibleSystem` and `FastSystem` terms a bit odd. Isn't it a bit context specific when they are fast / slow / convenient / inconvenient? Why not simply call them `SystemSOA` and `SystemAOS` then it is clear what they are. There could be a convenience constructor that can construct either, depending on a flag, with default probable being the AOS but I'm not sure about it.

src/interface.jl
```diff
@@ -4,74 +4,31 @@ using PeriodicTable
...
-#
-# The atom type itself
-#     - The atom interface is read-only (to allow as simple as possible implementation)
-#       Writability may be supported in derived or concrete types.
-#     - The inferface is only in Cartesian coordinates.
-#     - Has atom-specific defaults (i.e. assumes every entity represents an atom or ion)
-#
+Return the velocity of a particle `p`.
+"""
+velocity(p)::Union{Unitful.Velocity,Missing} = missing
```
cortner, line 17: could there be also a `momentum` function that call `velocity` and `mass` (and simply fails when there is no mass?)

src/interface.jl
```diff
@@ -81,82 +38,82 @@ struct DirichletZero <: BoundaryCondition end  # Dirichlet zero boundary (i.e. m
 struct Periodic <: BoundaryCondition end  # Periodic BCs
```
cortner, line 38: should this have a flag `true/false`?  Since the bc is implemented as a vector (or tuple) of BCs, you probably want each element of that vector to be e.g. a `Periodic()` . But now suppose you want a system that is open in the x, y direction and periodic in the z direction (e.g. dislocations). Then you need a different type `Open()` and you have a type instability.  But with `Periodic(true), Periodic(false)` there is no problem.

cortner: Personally, I don't really see how one needs anything other than a boolean here? I can see many other options such as reflecting, embedding, etc but those are not really relevant for the particle system per se, but rather for fields in the background, or for the embedding one would need additional structures anyhow into which to embed the System... 

But I don't see any harm in keeping the `Periodic` type.

src/interface.jl
```diff
@@ -81,82 +38,82 @@ struct DirichletZero <: BoundaryCondition end  # Dirichlet zero boundary (i.e. m
...
-#
-# The system type
-#     Again readonly.
-#
+"""
+    AbstractSystem{D,S}
+
+A `D`-dimensional system comprised of particles identified by type `S`.
+"""
+abstract type AbstractSystem{D,S} end
```
cortner, line 46: ah - there it is now. So if I want a particle system that has multiple element types then I can't have that? I agree that in 99% of the cases I probably don't want that because it will likely give me type instabilities. Just making sure this is really important enough to sacrifice the flexibility?

src/interface.jl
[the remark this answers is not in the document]
cortner: Is this intended to be implemented now during interface development? Then I think an error message is fine because it will flag that this is supposed to be implemented but isn't. But if it is an interface function that a user should implement, then I don't think the error message should be there.

cortner (commented):
looks good I think, just left a few fairly trivial comments.

src/implementation_soa.jl
rkurchin: 😆 hahaha that's what they were originally called and I changed them at @mfherbst's suggestion

src/interface.jl
rkurchin: Certainly...do we think that needs to be at this top level though?

src/interface.jl
rkurchin: I think the group of us has decent consensus that these belong here so I'm going to leave them in when I merge this, but I'll open an issue for further discussion of this point.

src/implementation_soa.jl
cortner: ah - sorry, that's what I get for not paying attention. Just don't have the time I'm afraid. I can't even quite put my finger on why I don't like those terms. I guess I prefer technically precise terms? Especially with an interface function a user will never have to worry about it until they know what they are doing and then they'll just go and read the docs?

src/implementation_aos.jl
rkurchin: Yes, that would probably be better. I was having trouble figuring out how to write it in a general enough way...will file an issue for this too.

src/interface.jl
cortner: it is so generic and universal that I think there is no harm introducing it? I don't have a strong view - you decide what you think is best.

src/interface.jl
cortner: ok

rkurchin:
I'm going to merge this. I've filed several issues to discuss lingering questions: #25, #26, #27, #28

src/implementation_aos.jl
cortner: The way I did it in JuLIP was to have a conversation function e.g. 
```julia
_convert_positions(X::Vector) = X 
_convert_positions(X::AbstractVector) = collect(X) 
_convert_positions(X::AbstractMatrix) = # convert into vector of SVectors 

_convert_bc(bc::NTuple{3, Bool}) = bc 
_convert_bc(bc::AbstractVector) = tuple(Bool.(bc)...)
```
and so forth. So this means that *ANY* sensible input will be automatically converted into the format needed internally.

cortner: ok, will repost in the issue.

PR #32, opened 2021-12-16
remove Base.show dispatch

rkurchin:
Having it there causes problems in downstream packages who have their own `show` methods but if they've only defined one of the `show` flavors and it gets invoked in REPL, it might fall back to this one, which is especially problematic if for any reason one of the interface functions it relies upon isn't dispatched...

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AtomsBase"
 uuid = "a963bdd2-2df7-4f54-a1ee-49d51e6be12a"
 authors = ["JuliaMolSim community"]
-version = "0.1.0"
+version = "0.1.1"
 
 [deps]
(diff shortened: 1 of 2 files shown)


mfherbst (commented):
Yeah I came across this when playing with AtomsBase in DFTK as well.

I'm ok with this solution, but I think there are two options to consider:
- Keep the function as in master and force downstream developers to do something about the extended printing (even if it just dispatches to the basic printing ... sort of **opt-out** approach.
- Keep the function but convert it to a non-standard interface (like call it `show_system` or something). This way downstream developers can easily **opt-in** to this default implementation.

jgreener64 (approved):
I ran into this too, favour removing for now.

If we do have a method I would favour opt-in over opt-out as some people don't mind how the types are shown.

PR #34, opened 2021-12-17
Refactor species and atoms

### LINKED ISSUE #33: Dynamic properties for `StaticAtom`
While playing a bit with AtomsBase support for DFTK I quickly ran into the issue that I need to attach extra "properties" to an atom for our internal bookkeeping (pseudos, magnetisation etc) and that the `element` which was stored inside the struct turned out to be not all that useful.

So how about we avoid storing the element at all (can always be recovered from the `atomic_symbol`) and allow appending/storing arbitrary properties using a `Dict`? Along the lines that got me thinking if might not be a better approach to just integrate with the standard Julia `properties` interface instead of having `atomic_property` at all. What I have in mind is something similar to https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/Energies.jl#L40 where the keys are the just "by convention". Naturally one could fill the Dict with some default data from a lookup into `elements` on atoms construction.

I don't think it makes sense to do this for some special properties like species, velocities, positions to be part of the Dict to ensure type stability on these, so those I would still keep these separate. 

Quick sketch what I have in mind:
```julia
struct Atom{D,L<:Unitful.Length}
    position::SVector{D,L}
    atomic_symbol::Symbol
    data::Dict{Symbol,Any}
end

function Atom(position, symbol::Union{Integer,AbstractString,Symbol})
    el = elements[symbol]
    data[:atomic_mass] = at.element.atomic_mass
    data[:atomic_number] = at.element.number    
    Atom(position, el.symbol, data)
end

Base.propertynames(at::Atom, private=false) = private ? append!(keys(data), fieldnames(at)) : keys(data)
Base.getproperty(at::Atom, x::Symbol) = x in keys(data) ?  getindex(data, x) : getfield(at, x)
Base.setproperty!(at::Atom, x::Symbol, v) = x in keys(data) ? setindex!(data, x, v) : setfield!(at, x, v)

# Maybe?
element(at::Atom) = elements[at.atomic_symbol]
```

Note that this is quite flexible and also future-proof. If we decide that a property is so important that it deserve to be a true field of `Atom` we just make it one and effectively "move it out" of the dict without any further code changes downstream.

Thoughts?

mfherbst:
Closes #33. As discussed today some refactoring. Status:

- [X] Refactoring done
- [x] Adapt tests
- [x] Tests for `Atom` struct
- [x] Adapt readme:
   * Adapt what is written
   * Rationale of `AtomView`
   * Rationale of `Atom` struct.
   * Rationale of `AtomicSystem`

Work-in-progress DFTK implementation based on this interface: https://github.com/JuliaMolSim/DFTK.jl/pull/558

Point to be discussed:
- I implemented a rather generic `AtomView` struct. This turned out to be basically impossible if `AbstractSystem` had a type-parameter on the species. So I decided to drop that. Instead we now have a function `species_type`.
- I made an `AtomicSystem` typedef, which is meant to be the flexible "goto" implementation people can use for parsing and returning data following our interface.


README.md
```diff
@@ -31,38 +44,136 @@ Functions that will work automatically:
...
+additional behavior depending on context.
 
 ### System state and properties
-The only required properties to be specified of the system is the species of each component of the system and the positions and velocities associated with each component. These are accessed through the functions `species`, `position`, and `velocity`, respectively. The default dispatch of these functions onto an `AbstractSystem` object is as a broadcast over it, which will "just work" provided the indexing/iteration interfaces have been implemented (see above) and the functions are defined on individual system components.
-
-As a concrete example, AtomsBase provides the `StaticAtom` type as this is anticipated to be a commonly needed representation. Its implementation looks as follows:
+The only required properties to be specified of the system is the species
+and implementations of standard functions accessing the properties of the species,
+currently `position`, `velocity`, `atomic_symbol`, `atomic_mass`, `atomic_number`,
+`n_dimensions`, `element`. The default dispatch of these functions onto an
```
rkurchin, line 63: strictly speaking, `n_dimensions`, while it is a property of both the species and the system, isn't accessed by a broadcast (since it has to match over the whole system anyway so the broadcast would be sort of redundant). Maybe rephrase this?

README.md
```diff
@@ -31,38 +44,136 @@ Functions that will work automatically:
...
+The file [src/fast_system.jl](src/fast_system.jl) contains an implementation of
+AtomsBase based on the struct-of-arrays approach. All species data is stored
+as plain arrays, but for convenience indexing of individual atoms is supported
+by a light-weight [`AtomView`](src/atomview.jl). See the implementation files
+as well as the tests for how these can be used.
+
+### Atoms and FlexibleSystem
+A flexible implementation of the interface is provided by the
+[`FlexibleSystem`]( src/flexible_system.jl) and the [`Atom`]( src/atom.jl) structs
+for representing atomic systems.
```
rkurchin, line 118: maybe add a brief remark somewhere around here (or perhaps above at the top of this H2 section) that while the interface is designed to be flexible to swapping things in and out, we hope that what we're provided (probably especially the `Atom`/`AtomView` types) can be the start of a relatively universal type? Just as a sort of nudge towards people thinking of adopting the interface to seriously consider adopting these types as well...

rkurchin (approved):
With the caveat that I haven't actually tried running any of this yet (and once I try applying it in my own stuff I may have some more to say, but I also likely won't get a chance to do that until next week), with the tiny exceptions of a few comments on the readme, this all looks awesome! 👍

README.md
mfherbst: agree.

README.md
mfherbst: Agree. I made that aspect a bit stronger. Better now?

README.md
```diff
@@ -31,38 +44,140 @@ Functions that will work automatically:
 * `n_dimensions`: returns `D`, the number of spatial dimensions of the system
```
jgreener64, line 44: The readme says `get_periodic` but this has changed to `periodicity`.

src/interface.jl
```diff
@@ -59,45 +63,90 @@ Return a vector of length `D` of `BoundaryCondition` objects, one for each direc
...
 function boundary_conditions end
 
+"""
+    species_type(::AbstractSystem)
+
+Return the type used to represent a species or atom.
+"""
+function species_type end
+
 periodicity(sys::AbstractSystem) = [isa(bc, Periodic) for bc in boundary_conditions(sys)]
```
jgreener64, line 73: Add a docstring.

src/interface.jl
```diff
@@ -59,45 +63,90 @@ Return a vector of length `D` of `BoundaryCondition` objects, one for each direc
...
+    n_dimensions(atom)
+
+Return number of dimensions.
+"""
 n_dimensions(::AbstractSystem{D}) where {D} = D
+# Note: Can't use ndims, because that is ndims(sys) == 1 (because of indexing interface)
 
 # indexing and iteration interface...need to implement getindex and length, here are default dispatches for others
 Base.size(s::AbstractSystem) = (length(s),)
 Base.setindex!(::AbstractSystem, ::Int) = error("AbstractSystem objects are not mutable.")
```
jgreener64, line 86: Does this need to throw a generic error? If not undefined (or perhaps mentioned in a comment) it would throw a `MethodError`. The error message as written may not be correct anyway, since someone could make a `mutable struct` system.

src/interface.jl
```diff
@@ -1,49 +1,53 @@
...
-    position(p)
+#
+# Species property accessors
+#
+"""The element corresponding to a species/atom (or missing)."""
+function element end
 
-Return the position of a particle `p`.
-"""
+"""The position of a species/atom."""
```
jgreener64, line 26: These docstrings are defined in two places, is there a benefit to them being here rather than just below?

jgreener64 (approved):
Looks good to me, I am fine with removing the species type parameter. I added some minor comments.

README.md
rkurchin: LGTM

src/interface.jl
mfherbst: Agreed. I removed the message completely (that's how StaticArrays does it for example).

src/atomview.jl
```diff
@@ -0,0 +1,15 @@
+#
+# A simple view datastructure for atoms of struct of array systems
+#
+export AtomView
+struct AtomView{FS <: AbstractSystem}
+    system::FS
+    index::Int
+end
+velocity(v::AtomView)      = missing
```
jrdegreeff, line 9: Why doesn't this call `velocity(v.system, v.index)`? For the built-in implementations that would be missing, but for other implementations of the interface that provide a velocity, this function won't work currently.

src/atom.jl
```diff
@@ -0,0 +1,77 @@
...
+export Atom, atomic_system, periodic_system, isolated_system
+
+struct Atom{D, L<:Unitful.Length, M<:Unitful.Mass}
+    position::SVector{D, L}
+    atomic_symbol::Symbol
+    atomic_number::Int
+    atomic_mass::M
+    data::Dict{Symbol, Any}  # Store arbitrary data about the atom.
+end
+velocity(::Atom)          = missing
```
jrdegreeff, line 13: What is the rationale for not including the velocity if we have gotten to the point of allowing for arbitrary data? For MD use cases, I basically have to copy this entire implementation into my own code and then change 2 lines. It's workable, but far from elegant and requires me to rewrite it every time we change the code here. Is there an argument against having an additional SVector for velocity in the default implementation seeing as it's part of the interface already?

jrdegreeff (commented):
See my comments on the lack of velocity support. If we could add that field to the `Atom` type, we could remove all the extra workarounds in the [Atomistic/AtomsBase integration code](https://github.com/cesmix-mit/Atomistic.jl/blob/main/src/integrations/ab_integration.jl) and make the whole workflow more streamlined (with that one change we can just use the implementations provided in `AtomsBase` itself without needing a custom implementation that users would have to know about).

Note that this would only impact the `flexible_system` implementation and not the `fast_system`. You can also just have a default velocity value in the existing constructor to not impact existing code if there is any (although I guess this PR is breaking anyway, so maybe we don't care about that detail).

src/atom.jl
mfherbst: Sounds good. I added a velocity field in the last commit. @jgreener64 any thoughts?

jgreener64: Seems fine to me.

PR #36, opened 2022-01-29
Initial version of docs from README

rashidrafeek:
I have just split the README file into separate sections and also added an  "API reference" section. Hopefully, this is fine as an initial version of the documentation.

Many functions and types don't have docs and therefore are not present here. Maybe these can be added later on? Have also fixed some minor typos in the README.


mfherbst (approved):
Thanks very much! Since you basically structured the readme into the docs, I suggest to avoid the duplication and make a very slim readme. E.g. just a few summarising sentences and a short example. Otherwise just refer to the docs. I'll get the GHA and deployment set up once this is merged.

rkurchin:
Yeah I would agree with @mfherbst about avoiding duplication, because it's just multiple things to maintain when things evolve down the line. LGTM otherwise! 🚀

jgreener64 (approved):
Looks good to me, I agree with the above comments.

One option I have liked in the past is to have the landing page of the docs (here index.md) be a symlink to the repo readme, since they should contain similar information. The content in index.md, with the badges added on top, looks suitable for the readme as is.

rashidrafeek:
Thanks for the comments. I have followed @jgreener64's suggestion and shortened the README to just include the contents in index.md and made index.md a link to README.md. Also, I have added installation instructions in the README.

jgreener64:
Looks good to me.

rkurchin:
Great, I'll go ahead and merge this then. Thanks again @rashidrafeek!

PR #38, opened 2022-03-15
Fix StackOverFlowError when constructing from Vector{Any}

mfherbst:


rkurchin:
This seems fine to me, shall we merge it in and trigger Registrator?

PR #41, opened 2022-03-24
Fix broken link to documentation

jameskermode:
The https://juliamolsim.github.io/AtomsBase.jl/stable link was broken, but I see https://juliamolsim.github.io/AtomsBase.jl/dev exists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ existing packages easy. We also provide a couple of
 that we envision to be broadly applicable.
 If features beyond these are required we
 encourage developers to open PRs or provide their own implementations.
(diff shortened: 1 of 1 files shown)


rkurchin:
Thanks! But this probably means we should fix the docs deployment so tagged versions actually create those `/stable` URL's...I think it was @jgreener64 who originally set these up IIRC, maybe he can twiddle with it?

jameskermode:
I thought this might be the case. Happy for this PR to be closed in that case.

jgreener64:
Thanks for flagging, https://github.com/JuliaMolSim/AtomsBase.jl/pull/42 should fix this.

PR #42, opened 2022-03-25
Fix docs deployment

jgreener64:
As described in https://github.com/JuliaMolSim/AtomsBase.jl/pull/41 the docs build is not triggering when TagBot runs.

I have added the appropriate repo secrets so this should solve the problem.


jgreener64:
This didn't cause the stable version docs to build - I think that will happen the next time a version is tagged.

mfherbst:
Indeed. Maybe we can rerun the action though. Let me try that.

jgreener64:
Thanks, that worked.

PR #47, opened 2022-05-13
Fixed broken link in README

chemicalfiend:
The current link in the README that goes to the Atomic Systems section in the docs doesn't work. I've changed it to the correct link.


jgreener64:
Thanks.

PR #52, opened 2022-08-17
Fix Atom() function call typo

jty-computation:
Atom function was referred to as Atoms in documentation.


rkurchin:
Thanks, @jty-computation! LGTM

PR #53, opened 2022-08-22
Store arbitrary data in Flexible System

JanHab:
Added a Dict in Flexible System to store arbitrary data about the system.
Additionally fixed in Atom that it is now possible to create 2D or 1D Atoms and not only 3D Atoms.


mfherbst (approved):
LGTM

PR #55, opened 2022-09-15
added interface for data

JanHab:
I added an interface to get the data of the flexible_system and of an atom with data(system), data(system, i) or data(Atom).


mfherbst:
Is this really needed? Is not `getproperty` and `propertynames` etc. sufficient?

My point is it is better to use the already existing property-interface from base rather than inventing a new one that does essentially the same thing.

rkurchin:
I agree with @mfherbst here – much better to use an existing Julia interface for this stuff since it already exists...unless there's an important use case that that doesn't cover? I don't see one, though...

JanHab:
I think you are write, thank you.

PR #56, opened 2022-10-16
Added routine `data(system)` to get arbitrary data of the system

JanHab:
Hello,
i have added a routine `data(system)` to get the data stored in the Dict data (in case of a FlexibleSystem).
For a FastSystem the routine will return "missing", similar to `velocity(system)`.

I have already created a similar PR once. 
Back then you argued that functions like `propertynames` and `getproperty` would be sufficient.
I was convinced that this could work. However, now i am pretty sure that a `data` routine is helpful in the case of an AtomsBase interface in another package.
With `getproperty` you can just call the property if its a direct property of the system.
However, if the property is stored in another property you run into problems.
Have a look on the AtomsBase Interface in [ExtXYZ.jl](https://github.com/libAtoms/ExtXYZ.jl/blob/master/src/atoms.jl?plain=1#L25-42).
There the struct `Atoms` has two members (atom_data and system_data).
This means that functions like `getproperty` and `propertynames` aren't sufficient in this case. It is possible to store extradata in this system, either in atom_data or system_data, but it is not possible to call them the same way as it is done in AtomsBase.

Because of points like this i think that an extra routine to get those properties would be good.
I am happy to hear any suggestions on this and on the name for the routine (not sure if `data` is the best decision hear)


rkurchin:
Maybe I'm missing something here, but couldn't you just define a dispatch of `getproperty` to work in this case? I would argue that's a bit more natural anyway. I think the constraint you're alluding to may apply in the case of [`getfield`](https://docs.julialang.org/en/v1/base/base/#Core.getfield) but (as the docstring itself illustrates) it shouldn't for [`getproperty`](https://docs.julialang.org/en/v1/base/base/#Base.getproperty)...

mfherbst:
As last time I'm a bit torn how to proceed here. I'm not a big fan of the `data` function still, but in discussions with @JanHab it emerged as one solution (but I'm not anymore convinced it is the best though).

My main issue with it is that it sort of exposes internal knowledge (in particular `data` sort of takes the "leftovers" that don't fit in any of the officially supported slots. So what happens if an officially supported slot is added. This is effectively a breaking change if we admit this PR as the property will now no longer end up in data. I don't think this is a good idea.

Let me elaborate on some of my further thoughts:

- I think the way the extra data is handled should be as transparent to the user as possible. This ensures that we can easily "promote" extra data to be part of the "proper" interface of AtomsBase as soon as we see need. E.g. in DFTK we read the `:magnetic_moment` property of the atoms using `hasproperty(atom, :magnetic_moment)` and the way the property handling works on the atom level ensures that this will continue to work if we ever promote this to be a required field same as `position` and `velocity`are now.
- We recently added system-wide custom properties. The same arguments as for atomic properties applies, I'd say.
- An aspect that is unclear to me is to what extent and how we should distinguish in the interface between system-wide and atom-wide (custom) properties. E.g. staying with the `getproperty` approach, should
   ```julia
   system.custom_atom_property
   ```
   return an array of the value of `custom_atom_property` of the custom atomic properties? What is then the equivalent to the two-argument version `position(system, index)`? Who wins in case there are conflicts (data of the same key present in both system and atoms)? What happens if only some atoms have a custom property? Of course we can define a custom `getproperty(system, symbol::Symbol, index::Integer)` function, but then calling it would not exactly be intuitive and breaks a little the reason for going with `getproperty` in the first place.
- Especially the last aspect lead me to think more along the line of a `getindex` (and thus enable something like `system[begin:end, :custom_atom_property]` to return the array of all custom atomic properties (and maybe error in case a property is not present) and  `system[1, :custom_atom_property]` to return only the value for the first atom. In some sense this supercharges the already existing syntax `system[1]` to return the first atom. Could however also create potential confusion, so I'm curious for other opinions. Also their might be caveats with teaching Julia how to properly iterate over a system that we need to figure out.
- If we go with getindex a la `system[1, :custom_atom_property]`, I suggest we additionally have `system[:custom_system_property]` and `atom[:custom_atom_property]` (such that `system[1, :custom_atom_property] == system[1][:custom_atom_property]`).
- Next question is obviously how does the user find out about the extra data. I think the parallel here is to go along the line of `Dict`s and use `haskey(system, :custom_system_property)`, `haskey(atom, :custom_atom_property)`, appropriate `keys(system)`, `keys(atom)`. To enable the clever mapping we usually do in the `AtomView` we would also something like `keys(system, index)` (where `keys(system, 1) == keys(system[1])` by default).
- Overall this points to perhaps revising the idea of using a property-based interface for accessing the custom keys and indeed even to drop it to avoid confusion (having too many possibilities to do something is also a bad idea IMHO).

I think adding this symbol-key-based interface into the AtomView appropriately should enable custom properties for ExtXYZ, but I might be mistaken. @JanHab what do you think?

JanHab:
Maybe the `data(system)` routine is not the best approach here.

I think adding the fields in the system-wide custom properties to the `propertynames` could work.
In my mind the best approach is, as mentioned above, to go with getindex:
`system[:custom_system_property]`, `system[i][:custom_atom_property]` and `system[1, :custom_atom_property]` to get all the custom set properties.
I would maybe change the order of the function arguments to `system[:custom_atom_property, i]`. In this case it would work similar to the interface where you, for example, call the position of a specific atom with `position(sys, i)`.

Maybe the best way would be to go with getindex and the following routines:
- `system[:custom_system_property]`
- `atom[:custom_atom_property`
- `system[:custom_atom_property, 1]` which would work the same as `system[1][:custom_atom_property]`
- `propertynames(system)`
- `hasproperty(system)` and `getproperty(system)`, which are already implemented

Going along the line of Dicts to find out about the extra data could work good. 
However, wouldn't it be similar to using properties? `propertynames(system)` would work the same as `keys(system)`?

I think both, the symbol-key-based interface and the property-based interface could work here, but maybe using `propertynames(system)`, `getproperty(system)` and `system[:custom_system_property]` is more intuitiv then `keys(system)`or `haskey(system, :custom_system_property)`, because the system is no Dict. However, both could work good.

I think adding the symbol-key-based interface into the AtomView could enable properties for ExtXYZ, but i also think that adding a property-based interface to ExtXYZ could work @mfherbst .

However, one could call the bounding box with `bounding_box(system)` or with `system[:box]`.
The bounding_box is saved as `box` internal, which could lead to confusion by using the properties, because of the two different names? I think it could be a good idea to rename `box` to `bounding_box` in this case?

What are your thoughts on this?

JanHab:
Hello, 
i have now added the `propertynames` and `getindex` routines as mentioned above.
What are your thoughts on this? Do you think we should take the property-based interface instead of the symbol-key-based interface? Or maybe a complete new approach?
I am happy to hear your thoughts on this.

JanHab:
After talking to @mfherbst we thought about replacing the property-based-interface with his suggestion of a key-based-interface.
I have tried to implement it so far and would be happy to read your thoughts about it.
We thought about dropping the property-based implementation to avoid confusion. So i completly replaced the `propertynames` for atom and FlexibleSystem with `keys`. Furthermore the `getproperty` and `hasproperty` routines are replaced with `getkey` and `haskey`.
What do you think about this?

rkurchin:
I don't have strong feelings about this, but I do think we should ideally choose and stick with one or the other, unless there's a compelling case to support both functionalities. What are the arguments for either option?

mfherbst:
@rkurchin To answer your question. I guess you recall the reasoning for properties: Nice intuitive interface that easily integrates with standard Julia.

The main problem is: How to access custom atom properties from the system side and do it in a way this works well for struct of arrays and arrays of structs. This implies that at the system level there needs to be a custom function taking both the atomic index and the desired atomic property. `getproperty` does not support such an interface in a way that it integrates with the familiar `system.property` syntax.

In contrast `getindex` seamlessly allows for that. E.g. `system[1]` gets the first atom, `system[1, :custom_property]` gets the `:custom_property` from the first atom, `system[:, :custom_property]` gets the custom properties from all atoms as a vector. In a consistent way `system[:custom_sys_property]` gets a :custom_sys_property` from the system, too.

mfherbst (commented):
@JanHab I think your implementation misses the `system[:, :custom_property]` syntax. Also I think you are missing functions for the `AtomView`. We should also add some documentation (I'm happy to do that), but otherwise it looks ok.

JanHab:
Yes, i could have forgotten functions for the `AtomView`.
I am not completly sure which are missing, so i am happy to see your suggestions here.

mfherbst:
>  I am not completly sure which are missing, so i am happy to see your suggestions here.

Oh sorry, you are right. But what you do for the fast system (i.e. `Base.keys(system::FastSystem) = propertynames(system)`) does not really make sense as e.g. `:atomic_masses` is an atomic property and not a system property. So there is certainly an inconsistency at this level.

I'll think about this some more and come back to you.

mfherbst:
@JanHab It turned out getting the key interface between fast systems and flexible systems consistent was a bit more effort than I originally thought, so I decided to extend this PR in #61. Please take a look and let me know if that's fine with you.

JanHab:
Your changes seem to be great
I think i can close this PR now.

PR #58, opened 2022-12-16
More detailed printing and extended docs

mfherbst:

PR #61, opened 2022-12-26
Key-based property interface

mfherbst:
Extension of the implementation discussed in #56. Applied on top of the changes made in #58, which should be merged first.

Todos:
- [ ] Adapt documentation


docs/src/tutorial.md
```diff
@@ -112,31 +135,68 @@ Property name       | Unit / Type        | Description
...
                            Atom(:H, [0, 0, 3.]u"bohr")],
-                           bounding_box, boundary_conditions)
+                           box, boundary_conditions)
 ````
 An update constructor for systems is supported as well (see [`AbstractSystem`](@ref)). For example
 ````@example system
 AbstractSystem(hydrogen; bounding_box=[[5.0, 0.0, 0.0], [0.0, 5.0, 0.0], [0.0, 0.0, 5.0]]u"Å")
 ````
-Note that `FlexibleSystem( ... )` would have worked as well in this example (since we are
+To update the atomic composition of the system, this function supports an `atoms` (or `particles`)
```
rkurchin, line 161: My local (inside VS Code) Markdown renderer is showing these line breaks. I can't remember if GH Markdown makes them or does thing the LaTeX way where it won't make a line break unless the source has two. [This discussion](https://github.com/community/community/discussions/10981) seems to suggest the line breaks here should be fine, but let's keep an eye out...

src/atom.jl
```diff
@@ -55,16 +56,15 @@ end
 function Atom(id::AtomId, position::AbstractVector, velocity::Missing; kwargs...)
     Atom(id, position; kwargs...)
 end
+function Atom(; atomic_symbol, position, velocity=zeros(length(position))u"bohr/s", kwargs...)
```
rkurchin, line 59: Why does this one set the default velocity to zero when the one above sets it to missing?

rkurchin (commented):
Generally looks good – I put a few comments (I accidentally did them outside of this "review" so apologies if that muddles anything) raising some very minor things.

docs/src/tutorial.md
mfherbst: In GH-flavoured  markdown linebreaks are ignored. Documenter does this as well.

src/atom.jl
mfherbst: Well actually the first catches the case where it is equal to `missing` and changes that to zero. I can make that more explicit.

In fact in my usage of AtomsBase the missing business caused quite some trouble (like duplicated code paths just to cover for the "missing" case). I'm almost convinced the better option would be to drop the missing idea (and just return zeros instead).

PR #64, opened 2023-01-16
add CITATION.cff file

rkurchin:
I'm writing a manuscript where I want to reference AtomsBase, so I thought it would be nice to add this to the repository to make it easy for anyone else wanting to do so in the future. We could also consider generating a DOI for the package at some point (maybe when we decide to tag v1.0 or something)...

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,15 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
+authors:
+- family-names: "Herbst"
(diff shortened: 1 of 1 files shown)


rkurchin:
@mfherbst LMK if you think we should add more authors; you and I are definitely the majority by commits/LOC at the moment but I'm not opposed either...

mfherbst:
Sounds fair. The gap is quite large to the next lot of people. I'm also not opposed to adding anyone else at any point in the future.

Issue #29, opened 2021-11-18
Should slicing work? If so, what should it do?


rkurchin:
Currently, you can only use single indices. Intuitively, I would say slicing should return a new system with the same bounding box, boundary conditions, etc. but only containing the sliced indices. However, one could also imagine it returning a list of the actual particle objects referenced by the indices...also, either of these becomes a bit fraught in the case of a lattice with Cartesian indexing because the bounding box is more "intimately" connected to the specific particles/sites...so maybe not supporting slicing at the top level is fine?

jgreener64:
I would lean towards it returning a list of the actual particle objects referenced by the indices; this means it would be possible to define a default at the top level, which I think is difficult in the first case.

If we added
```julia
Base.getindex(s::AbstractSystem, r::Union{UnitRange, AbstractArray}) = getindex.((s,), r)
```
that would provide this slicing behaviour, e.g. for the test example
```julia
julia> aos[1:2]
2-element Vector{StaticAtom{3, Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}}:
 StaticAtom: C
 StaticAtom: C

julia> aos[[1, 2]]
2-element Vector{StaticAtom{3, Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}}:
 StaticAtom: C
 StaticAtom: C
```

I agree that non-standard indexing will complicate this a bit but the array case above would still work with an array of more complex indices.

Agree it's borderline whether it is worth supporting at the top level though.

Kolaru:
I agree that slicing should return a vector of particle object. I think that this kind of advanced indexing is the natural way of querying the particles.

On simple example is getting all hydrogen atoms in the system. It could look something like the following
```julia
hydrogens = system[element.(system) .== :H]
```
(provided `element` is defined for the particle in the system). This requires indexing with a boolean array rather than slicing, but I believe it follows the same global idea.

mfherbst:
With @rashidrafeek's PR the discussed features should work now. Closing this for now.

PR #67, opened 2023-02-07
Slicing support

rashidrafeek:
Now with the key-based property interface we can do `system[:, :property]` to get the corresponding `:property` for all atoms. Wouldn't it also make sense to support `system[rng, :property]` for a subset, `rng` of atom indices. Also, from the discussion in #29, I think returning a list of particle objects for slicing is more intuitive and useful. In this PR, I have implemented both of these, i.e., returning vector of particles objects on slicing AbstractSystems and the corresponding extension to the key-based property interface.

For example, with this, we can do `system[atomic_symbol(system) .== :H]` to get all H atoms.

I am just putting this here as a working version to show the usefulness of this feature. Feel free to close this in case we think slicing need not be supported.

### DIFF
--- a/src/fast_system.jl
+++ b/src/fast_system.jl
@@ -75,5 +75,7 @@ Base.keys(::FastSystem) = (:bounding_box, :boundary_conditions)
 # Atom and atom property access
 atomkeys(::FastSystem) = (:position, :atomic_symbol, :atomic_number, :atomic_mass)
 hasatomkey(system::FastSystem, x::Symbol) = x in atomkeys(system)
-Base.getindex(system::FastSystem, i::Union{Integer,AbstractVector}, x::Symbol) = getfield(system, x)[i]
+function Base.getindex(system::FastSystem, i::Union{Integer,AbstractVector}, x::Symbol)
+    getfield(system, x)[i]
+end
 Base.getindex(system::FastSystem, ::Colon, x::Symbol) = getfield(system, x)
--- a/src/fast_system.jl
+++ b/src/fast_system.jl
@@ -48,6 +48,8 @@ Base.size(sys::FastSystem)           = size(sys.position)
 
 species_type(::FS) where {FS <: FastSystem} = AtomView{FS}
 Base.getindex(sys::FastSystem, i::Integer)  = AtomView(sys, i)
+Base.getindex(sys::FastSystem, rng::AbstractVector)  = [sys[i] for i in rng]
+Base.getindex(sys::FastSystem, rng::Colon)  = collect(sys)
 
 position(s::FastSystem)       = s.position
 atomic_symbol(s::FastSystem)  = s.atomic_symbol
@@ -75,5 +77,5 @@ Base.keys(::FastSystem) = (:bounding_box, :boundary_conditions)
 # Atom and atom property access
 atomkeys(::FastSystem) = (:position, :atomic_symbol, :atomic_number, :atomic_mass)
 hasatomkey(system::FastSystem, x::Symbol) = x in atomkeys(system)
-Base.getindex(system::FastSystem, i::Integer, x::Symbol) = getfield(system, x)[i]
+Base.getindex(system::FastSystem, i::Union{Integer,AbstractVector}, x::Symbol) = getfield(system, x)[i]
 Base.getindex(system::FastSystem, ::Colon, x::Symbol) = getfield(system, x)
--- a/src/fast_system.jl
+++ b/src/fast_system.jl
@@ -48,9 +48,6 @@ Base.size(sys::FastSystem)           = size(sys.position)
 
 species_type(::FS) where {FS <: FastSystem} = AtomView{FS}
 Base.getindex(sys::FastSystem, i::Integer)  = AtomView(sys, i)
-Base.getindex(sys::FastSystem, rng::AbstractVector)  = [sys[i] for i in rng]
-Base.getindex(sys::FastSystem, rng::AbstractVector{<:Bool})  = sys[(1:length(sys))[rng]]
-Base.getindex(sys::FastSystem, rng::Colon)  = collect(sys)
 
 position(s::FastSystem)       = s.position
 atomic_symbol(s::FastSystem)  = s.atomic_symbol
--- a/src/fast_system.jl
+++ b/src/fast_system.jl
@@ -49,6 +49,7 @@ Base.size(sys::FastSystem)           = size(sys.position)
 species_type(::FS) where {FS <: FastSystem} = AtomView{FS}
 Base.getindex(sys::FastSystem, i::Integer)  = AtomView(sys, i)
 Base.getindex(sys::FastSystem, rng::AbstractVector)  = [sys[i] for i in rng]
+Base.getindex(sys::FastSystem, rng::AbstractVector{<:Bool})  = sys[(1:length(sys))[rng]]
 Base.getindex(sys::FastSystem, rng::Colon)  = collect(sys)
 
 position(s::FastSystem)       = s.position
--- a/src/flexible_system.jl
+++ b/src/flexible_system.jl
@@ -27,7 +27,9 @@ Base.keys(system::FlexibleSystem) = (:bounding_box, :boundary_conditions, keys(s
 # Atom and atom property access
 Base.getindex(system::FlexibleSystem, i::Integer) = system.particles[i]
 Base.getindex(system::FlexibleSystem, i::Integer, x::Symbol) = system.particles[i][x]
-Base.getindex(system::FlexibleSystem, rng::AbstractVector, x::Symbol) = [at[x] for at in system.particles[rng]]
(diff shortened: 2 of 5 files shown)


rkurchin:
Cool! I'd like to see others' thoughts and perhaps discuss a bit, but I'm certainly preliminarily in favor of merging this (or something like it) in!

src/flexible_system.jl
```diff
@@ -25,8 +25,9 @@ end
 Base.keys(system::FlexibleSystem) = (:bounding_box, :boundary_conditions, keys(system.data)...)
 
 # Atom and atom property access
-Base.getindex(system::FlexibleSystem, i::Integer) = system.particles[i]
+Base.getindex(system::FlexibleSystem, i::Union{Integer, AbstractVector, Colon}) = system.particles[i]
```
mfherbst, line 28: Is `AbstractVector` maybe too general? Can you check in base whether `getindex` with anything but integers is even allowed in the code?

src/fast_system.jl
```diff
@@ -48,6 +48,8 @@ Base.size(sys::FastSystem)           = size(sys.position)
 
 species_type(::FS) where {FS <: FastSystem} = AtomView{FS}
 Base.getindex(sys::FastSystem, i::Integer)  = AtomView(sys, i)
+Base.getindex(sys::FastSystem, rng::AbstractVector)  = [sys[i] for i in rng]
+Base.getindex(sys::FastSystem, rng::Colon)  = collect(sys)
```
mfherbst, line 52: Indeed that's a way this could be solved. Another option could be to actually make a new system, but just with the respective atoms selected by the mask. Both could be useful and I see pros and cons in both directions. What does ASE do? What do others think?

mfherbst (commented):
Nice idea. Certainly in favour of having something like this.

src/flexible_system.jl
rashidrafeek: Here, `system.particles` is an AbstractVector and Base does [allow indexing](https://docs.julialang.org/en/v1/manual/arrays/#man-supported-index-types) any vector with a vector of indices or array of Booleans (logical indexing) like below:
```julia
julia> a = [1,2,3]
3-element Vector{Int64}:
 1
 2
 3

julia> a[1:2]
2-element Vector{Int64}:
 1
 2

julia> a[[1,2]]
2-element Vector{Int64}:
 1
 2

julia> a[[false,true,true]]
2-element Vector{Int64}:
 2
 3
```
These are the types of ranges I intended to implement (see the tests). Maybe `AbstractVector` is too general? I can make it more specific for these cases. Like `Union{AbstractRange, Colon, Vector{Int}, Vector{Bool}} `.

src/fast_system.jl
rashidrafeek: Making a new system was suggested initially by @rkurchin in #29. And @jgreener64 and @Kolaru, suggested returning a vector of particle objects. I went by the latter as that is trivial to implement and made more sense to me if we think of AbstractSystems as "a collection of atoms". But the former would also make sense and would be a way to remove some atoms from a system. The main issue with it would be of having a general way to create a new AbstractSystem of the same type (FlexibleSystem or FastSystem). But maybe this is possible now with the functionality present currently?

In terms of other packages, ASE creates a new system while pymatgen return a vector of particle objects. So there are previous examples of both behaviours.

src/flexible_system.jl
rashidrafeek: Tests for logical indexing were not present which I have added now.

src/fast_system.jl
jgreener64: I feel that returning a vector of particles is cleaner. It also allows more creative indexing, for example:
```julia
julia> a = [1,2,3]
3-element Vector{Int64}:
 1
 2
 3

julia> a[[3 3; 2 2]]
2×2 Matrix{Int64}:
 3  3
 2  2
```
So more generally the shape of the output matches the shape of the desired indices. This could have uses in the atomic context but it is non-trivial if the result has to be put into a new system.

Perhaps we could use `filter` for returning sub-systems.

src/flexible_system.jl
jgreener64: I would go even more general than `AbstractVector`, see an example use of a 2D array in my other comment.

jgreener64: Also, I would avoid concrete types like `Vector{Int}` as they exclude for example `CuArray`s.

Is the type annotation required at all? The error thrown from `system.particles[i]` may be quite readable in incompatible cases.

jgreener64:
Looks good, definitely in favour of some version of this.

src/flexible_system.jl
mfherbst: > Is the type annotation required at all?

In the concrete case of the fast / flexible system this is probably debatable, but let's just keep in mind that we are also defining an abstract interface and we should think a little of the possible burden it could be for downstream packages to allow "anything". Probably this is not too bad as the fallback @rashidrafeek provides should work in most cases. Still, we should clearly state in the docs what is the expected minimum in terms of support.

src/fast_system.jl
mfherbst: Ok, I agree with the idea to have this return an array and rather have a function `filter` or similar to return a subsystem.

src/flexible_system.jl
rashidrafeek: How about defining the following for all `AbstractSystems`:
```julia
Base.getindex(s::AbstractSystem, rng::AbstractArray) = getindex.(Ref(s), rng)
Base.getindex(s::AbstractSystem, ::Colon) = collect(s)
Base.getindex(s::AbstractSystem, r::AbstractVector{<:Bool}) = s[ (firstindex(s):lastindex(s))[r] ]
```
Then concrete types have to just implement `getindex` for `Integer` and all other cases automatically work, even 2 D arrays. This would simplify the code and the `getindex` I defined separately for FlexibleSystem and FastSystem in this PR can be removed.

rashidrafeek: I have updated the code to the above implementation in case someone needs to test this. Also have added tests for indexing with 2D arrays.

src/flexible_system.jl
```diff
@@ -27,6 +27,7 @@ Base.keys(system::FlexibleSystem) = (:bounding_box, :boundary_conditions, keys(s
 # Atom and atom property access
 Base.getindex(system::FlexibleSystem, i::Integer) = system.particles[i]
 Base.getindex(system::FlexibleSystem, i::Integer, x::Symbol) = system.particles[i][x]
+Base.getindex(system::FlexibleSystem, rng::AbstractVector, x::Symbol) = [at[x] for at in system.particles[rng]]
```
mfherbst, line 30: Maybe for consistency also call this `i` instead of `rng` (rng reminds me of random number generator).

mfherbst: Also I think I would do a multi-line function here to avoid the long line.

src/interface.jl
```diff
@@ -75,6 +75,9 @@ Base.lastindex(s::AbstractSystem) = length(s)
 Base.iterate(sys::AbstractSystem, state = firstindex(sys)) =
     (firstindex(sys) <= state <= length(sys)) ? (@inbounds sys[state], state + 1) : nothing
 Base.eltype(sys::AbstractSystem) = species_type(sys)
+Base.getindex(s::AbstractSystem, rng::AbstractArray) = getindex.(Ref(s), rng)
+Base.getindex(s::AbstractSystem, ::Colon) = collect(s)
+Base.getindex(s::AbstractSystem, r::AbstractVector{<:Bool}) = s[ (firstindex(s):lastindex(s))[r] ]
```
mfherbst, line 80: Maybe multiline? Also `<: Bool` can be just `Bool` as `Bool` is a concrete type.

test/atom.jl
```diff
@@ -58,9 +58,17 @@ using Test
...
         @test system[3] == atoms[3]
+        @test system[1:2] == [system[1], system[2]]
+        @test system[[1,3]] == [system[1], system[3]]
+        @test system[[1 3; 2 1]] == [system[1] system[3]; system[2] system[1]]
+        @test system[:] == [system[1], system[2], system[3]]
+        @test system[[false,false,true]] == [system[3]]
         @test system[3, :dat] == 3.0
         @test system[2, :atomic_symbol] == :C
+        @test system[1:2, :atomic_number] == [14, 6]
+        @test system[[1,3], :atomic_number] == [14, 1]
```
mfherbst, line 69: I'd put a space after each comma.

mfherbst (commented):
LGTM, just found a few nits.

src/flexible_system.jl
rashidrafeek: Done.

src/interface.jl
rashidrafeek: Done.

test/atom.jl
rashidrafeek: Fixed.

mfherbst (approved):

PR #68, opened 2023-03-21
Make AtomView rely on struct-of-array format

Liozou:
Hi! Here is a proposal to simplify a little bit the definition of struct-of-array systems, like `FastSystem`.

Currently, the two following definitions are required for each property like `position`:
```julia
position(s::FastSystem) = s.position
position(s::FastSystem, i) = s.position[i]
```

Since they are quite redundant, I suggest not requiring the latter. Instead, a `position(system, i)` call would be intercepted by the already existing fallback
https://github.com/JuliaMolSim/AtomsBase.jl/blob/092756182c37a862989da88ebb3fea6c14154388/src/interface.jl#L100
which can be made to work simply by tweaking the definitions of `position(v::AtomView)`. The emitted code should be identical in the end.

Also add some documentation for `AtomView` (since it is already exported and mentioned in the docs).

### DIFF
--- a/src/fast_system.jl
+++ b/src/fast_system.jl
@@ -57,8 +57,12 @@ velocity(::FastSystem)        = missing
 
 # System property access
 function Base.getindex(system::FastSystem, x::Symbol)
-    x === :bounding_box && return bounding_box(system)
-    x === :boundary_conditions && return boundary_conditions(system)
+    if x === :bounding_box
+        return bounding_box(system)
+    end
+    if x === :boundary_conditions
+        return boundary_conditions(system)
+    end
     throw(KeyError(x))
 end
 Base.haskey(::FastSystem, x::Symbol) = x in (:bounding_box, :boundary_conditions)
--- a/src/fast_system.jl
+++ b/src/fast_system.jl
@@ -55,19 +55,11 @@ atomic_number(s::FastSystem)  = s.atomic_number
 atomic_mass(s::FastSystem)    = s.atomic_mass
 velocity(::FastSystem)        = missing
 
-position(s::FastSystem, i)      = s.position[i]
-atomic_symbol(s::FastSystem, i) = s.atomic_symbol[i]
-atomic_number(s::FastSystem, i) = s.atomic_number[i]
-atomic_mass(s::FastSystem, i)   = s.atomic_mass[i]
-velocity(::FastSystem, i)       = missing
-
 # System property access
 function Base.getindex(system::FastSystem, x::Symbol)
-    if x in (:bounding_box, :boundary_conditions)
-        getfield(system, x)
-    else
-        throw(KeyError("Key $x not found"))
-    end
+    x === :bounding_box && return bounding_box(system)
+    x === :boundary_conditions && return boundary_conditions(system)
+    throw(KeyError(x))
 end
 Base.haskey(::FastSystem, x::Symbol) = x in (:bounding_box, :boundary_conditions)
 Base.keys(::FastSystem) = (:bounding_box, :boundary_conditions)
--- a/src/fast_system.jl
+++ b/src/fast_system.jl
@@ -58,12 +58,12 @@ velocity(::FastSystem)        = missing
 # System property access
 function Base.getindex(system::FastSystem, x::Symbol)
     if x === :bounding_box
-        return bounding_box(system)
+        bounding_box(system)
+    elseif x === :boundary_conditions
+        boundary_conditions(system)
+    else
+        throw(KeyError(x))
     end
-    if x === :boundary_conditions
-        return boundary_conditions(system)
-    end
-    throw(KeyError(x))
 end
 Base.haskey(::FastSystem, x::Symbol) = x in (:bounding_box, :boundary_conditions)
 Base.keys(::FastSystem) = (:bounding_box, :boundary_conditions)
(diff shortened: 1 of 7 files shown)


src/fast_system.jl
```diff
@@ -55,19 +55,11 @@ atomic_number(s::FastSystem)  = s.atomic_number
...
 
-position(s::FastSystem, i)      = s.position[i]
-atomic_symbol(s::FastSystem, i) = s.atomic_symbol[i]
-atomic_number(s::FastSystem, i) = s.atomic_number[i]
-atomic_mass(s::FastSystem, i)   = s.atomic_mass[i]
-velocity(::FastSystem, i)       = missing
-
 # System property access
 function Base.getindex(system::FastSystem, x::Symbol)
-    if x in (:bounding_box, :boundary_conditions)
```
mfherbst, line 66: That's a bit too smart for my taste :smile:. I'd use explicit if branches.

test/printing.jl
```diff
@@ -13,12 +13,12 @@ using Test
 
     flexible_system = periodic_system(atoms, box; fractional=true, data=-12)
     @test repr(flexible_system) == """
-    FlexibleSystem(CSi, periodic = TTT, bounding_box = [[10.0, 0.0, 0.0], [0.0, 5.0, 0.0], [0.0, 0.0, 7.0]]u"Å")"""
```
mfherbst, line 16: Interesting. What's happening here?

mfherbst (approved):
Very nice. I think we should've done this in the first place.

test/printing.jl
Liozou: For some reason this test failed because the exact encoding of the angstrom character differed... Turns out this is simply https://github.com/PainterQubits/Unitful.jl/pull/603

src/fast_system.jl
Liozou: Ah sure! I changed this part because for my specific use case `bounding_box` is a field but not `boundary_conditions`, and writing it this way instead of the previous way made it clearer to me. But explicit branches are probably even clearer, I'll fix this right now

src/fast_system.jl
```diff
@@ -55,19 +55,15 @@ atomic_number(s::FastSystem)  = s.atomic_number
 atomic_mass(s::FastSystem)    = s.atomic_mass
 velocity(::FastSystem)        = missing
 
-position(s::FastSystem, i)      = s.position[i]
```
mfherbst, line 58: Sorry just came to my mind: Can you check this has no impact on type stability?

src/fast_system.jl
mfherbst: Also for consistency: Could youmake the same changes for the flexible system?

Liozou: Done in https://github.com/JuliaMolSim/AtomsBase.jl/pull/68/commits/0bbb428af8ac90803f4def98d382d1d297be96ba

src/fast_system.jl
Liozou: Good point: I added some tests for type stability in https://github.com/JuliaMolSim/AtomsBase.jl/pull/68/commits/bd269102ed48ae40037a88b656657eea0427b1fb

test/fast_system.jl
```diff
@@ -39,6 +42,12 @@ using PeriodicTable
         :atomic_mass => atomic_mass(atoms[1]),
     ]
 
+    # check type stability
+    get_b_vector(syst) = bounding_box(syst)[2]
+    @test @inferred(get_b_vector(system)) === SVector{3}([0.0, 1.0, 0.0]u"m")
+    @test @inferred(position(system, 1)) === SVector{3}([0.25, 0.25, 0.25]u"m")
```
mfherbst, line 48: I'd not use `===` here. `==` should be sufficient.

Liozou: Indeed, but since we are testing for type stability, I thought we might as well check that there is no unexpected internal unit conversion done by Unitful?

mfherbst: I'm not sure, but for `isbits` types `==` and `===` are probably exactly the same, so it's more a matter of philosophy ... and there I'd argue that I'm not in comparing the bit-level representation here. I'd more care about the values being equal (thus `==`).

Liozou: I don't really care about the bit-level representation either, but I thought it would be good to check that the unit remains that which was given, and `==` is not enough for that purpose since
```julia
julia> SVector{3}([0.25, 0.25, 0.25]u"m") == SVector{3}([25.0, 25.0, 25.0]u"cm")
true

julia> SVector{3}([0.25, 0.25, 0.25]u"m") === SVector{3}([25.0, 25.0, 25.0]u"cm")
false
```
But actually we may not care much about that either... So, sure, I'll revert to `==`.

Liozou: https://github.com/JuliaMolSim/AtomsBase.jl/pull/68/commits/40448744a5213250069f4db5b888c72868080013

mfherbst: ah ok, that distinction I did not think about, but makes sense to have the two versions behave differently. Well too be fair it probably does not matter if internally something is converted, provided that the actual value is still the same.

Liozou: Yes, it's not really important actually, as long as it's type stable it shouldn't be an issue.

Liozou:
Great, thanks for the review!

PR #69, opened 2023-03-25
Add AtomsBaseTesting subpackage

mfherbst:
Adds a small subpackage, which can generate standard test cases and test equality between AtomsBase systems. Already used in ExtXYZ, [AtomsIO](https://github.com/mfherbst/AtomsIO.jl/blob/master/test/common.jl) and [ASEconnect](https://github.com/mfherbst/ASEconvert.jl/blob/master/test/common.jl) for testing (with code duplicated across these three packages). Clearly this has broader use, so this will allow registering a separate `AtomsBaseTesting` package.

### DIFF
--- a/lib/AtomsBaseTesting/src/AtomsBaseTesting.jl
+++ b/lib/AtomsBaseTesting/src/AtomsBaseTesting.jl
@@ -53,6 +53,7 @@ function test_approx_eq(s::AbstractSystem, t::AbstractSystem;
         if hasatomkey(s, prop) != hasatomkey(t, prop)
             println("hashatomkey mismatch for $prop")
             @test hasatomkey(s, prop) == hasatomkey(t, prop)
+            continue
         end
         for (at_s, at_t) in zip(s, t)
             @test haskey(at_s, prop) == haskey(at_t, prop)
@@ -77,6 +78,7 @@ function test_approx_eq(s::AbstractSystem, t::AbstractSystem;
         if haskey(s, prop) != haskey(t, prop)
             println("haskey mismatch for $prop")
             @test haskey(s, prop) == haskey(t, prop)
+            continue
         end
 
         if s[prop] isa Quantity
@@ -97,7 +99,6 @@ and specific standard keys can be ignored using `drop_atprop` and `drop_sysprop`
 """
 function make_test_system(D=3; drop_atprop=Symbol[], drop_sysprop=Symbol[],
                           extra_atprop=(; ), extra_sysprop=(; ), cellmatrix=:full)
-    # TODO Should be moved to AtomsBase
     @assert D == 3
     n_atoms = 5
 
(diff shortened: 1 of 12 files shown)


lib/AtomsBaseTesting/test/runtests.jl
```diff
@@ -0,0 +1,35 @@
+using AtomsBaseTesting
+using Test
+using LinearAlgebra
+using AtomsBase
+
+@testset "AtomsBaseTesting.jl" begin
+    @testset "Run generation and testing code" begin
+        case = make_test_system()
+        test_approx_eq(case.system, case.system)
```
jgreener64, line 9: Would be good to test a case that fails, and maybe a case that succeeds/fails based on tolerance.

I realise that testing the negative is non-trivial with the current `@test` setup since it will cause the outer tests to fail. Perhaps a case for `try`/`catch` or `@test_broken`? Or changing `test_approx_eq` to return a Boolean? That may make it less useful as it won't show where the difference is.

lib/AtomsBaseTesting/src/AtomsBaseTesting.jl
```diff
@@ -0,0 +1,161 @@
...
+
+
+"""
+Setup a standard test system using some random data and supply the data to the caller.
+Extra atomic or system properties can be specified using `extra_atprop` and `extra_sysprop`
+and specific standard keys can be ignored using `drop_atprop` and `drop_sysprop`.
+"""
+function make_test_system(D=3; drop_atprop=Symbol[], drop_sysprop=Symbol[],
+                          extra_atprop=(; ), extra_sysprop=(; ), cellmatrix=:full)
+    # TODO Should be moved to AtomsBase
```
jgreener64, line 100: Is this suggesting that this function should be in AtomsBase long term?

jgreener64 (commented):
Looks like a useful feature. What is the value of having this as a separate package? It depends on AtomsBase so isn't any more lightweight.

mfherbst:
Thanks for the review.

> What is the value of having this as a separate package? 

The issue is it depends on `Tests` and maybe in the future potential test data, which packages using `AtomsBase` don't need. So essentially this is done to keep `AtomsBase` and not the reverse :smile:

lib/AtomsBaseTesting/test/runtests.jl
mfherbst: Good point. I'm certainly against changing `test_approx_eq` to only return a bool for the reason you write: That's not super instructive. My main idea with this was to run the largest chunk of the code and catch trivial issues. Maybe test error can be caught though. I'll think about it.

lib/AtomsBaseTesting/src/AtomsBaseTesting.jl
```diff
@@ -0,0 +1,161 @@
...
+    @assert D == 3
```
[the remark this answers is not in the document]
mfherbst: In theory the whole code should work for 1D and 2D as well, but that's not consistently written / tested as of now, so it basically assumes D == 3, hence I've put an assertion. I think given that this will be used in a testing environment that's cleaner than using an exception (which one would rather expect to come from the tested code, not the test setup).

jgreener64 (approved):

PR #72, opened 2023-05-31
Nicer error message for string-based element lookup

### LINKED ISSUE #71: Element can't be found when using String for AtomID
https://github.com/JuliaMolSim/AtomsBase.jl/blob/652f9166fc3c1c8c04f76b4c99e28aa31db3002a/src/interface.jl#L91

MWE:
```
Atom("Si", zeros(3)u"\AA")
```

mfherbst:
Fixes #71

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -88,7 +88,24 @@ end
 #
 
 """The element corresponding to a species/atom (or missing)."""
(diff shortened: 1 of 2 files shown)


louisponet:
I probably like this most out of the 3 proposals in the issue! Too much magic almost never feels good imo.

jgreener64:
Looks good.

PR #80, opened 2023-08-21
add AutoBZCore to readme

lxvm:
Hi,
I recently added an [extension to AutoBZCore.jl](https://github.com/lxvm/AutoBZCore.jl/blob/main/ext/AtomsBaseExt.jl) to interface with AtomsBase.jl systems. I thought adding the package to the readme would help improve visibility if it seems appropriate.
Cheers,
Lorenzo


mfherbst:
Sure, the more the merrier.

Issue #84, opened 2023-09-25
Energies, Forces, Etc


cortner:
I propose that we stop ~dragging our feet~ sleeping on this and implement prototypes for 
```
potential_energy
forces
virial
```
We can add more if desired e.g. 
```
stress
hessian
```
come to mind. `potential_energy` could also be called just `energy`, I don't mind either way.

This will make an ultra-lightweight interface. If needed it can be moved out of `AtomsBase` into a separate package, but I see no harm having it here for now. If `Molly` were the only thing that does simulations then it could live there, but it's not. 

If people agree please `:thumbsup` and  I'll make a PR.

CC @jgreener64 @rkurchin  @mfherbst  @tjjarvinen 

**UPDATE:**  after discussion the current proposal is the following: 
```julia
potential_energy(system, calculator; kwargs...)
forces(system, calculator; kwargs...)
forces!(F, system, calculator; kwargs...)
virial(system, calculator; kwargs...)
energy_forces_virial(system, calculator; kwargs...)
energy_forces_virial!(F, system, calculator; kwargs...)
```

jgreener64:
I can see what the output of these functions is, but what is the input? I guess the whole system?

Currently in Molly the first three functions take in the system, the neighbour list and the number of threads but this can be reworked to be just the system.

I prefer `potential_energy` to `energy`.

cortner:
My suggestion is 
```
potential_energy(atoms, calc)::Real
forces(atoms, calc)::AbstractVector{<: SVector}
virial(atoms, calc)::AbstractMatrix
```
but in some ways, this is not that important. it's really easy to re-interpret arrays or change around arguments. First I want shared prototypes. 

Of course if we can agree and document what they should return that's even better.

cortner:
There could also be an 
```
energy_forces_virial(atoms, calc)::Tuple{...}
```
to compute everything at the same time, that way we don't need to cache things.

jgreener64:
What is `calc`?

cortner:
calculator (ASE language)

tjjarvinen:
This is a good idea when there starts to be more packages that use forces. With this you only need to implement one interface for a "calculator" package for it to work everywhere.

I also have some suggestions:

Add kwargs to functions calls 

```julia
potential_energy(atoms, calc; kwargs...)::Number
forces(atoms, calc; kwargs...)::AbstractVector{<:AbstractVector}
[...]

PR #85, opened 2023-09-27
[WIP] Prototypes for Energy, Forces, Virials

cortner:
This PR goes with #84  -  For the time being I propose to do no more than 

- [x] define minimal prototypes as in the first commit.  
- [ ] document how they are intended to be used.  (to be added to the docs)
- [ ] possibly provide tests that check for correctness of output, not clear this is needed for a first PR but could be nice to have.
- [ ] some default fallbacks; cf Teemu's post below

For now I am leaving out the general `calculate` prototype and will add it only if there is at least partial agreement on that.

### DIFF
--- a/src/calculators.jl
+++ b/src/calculators.jl
@@ -1,0 +1,19 @@
+
+
+function potential_energy end 
+
+function forces end 
+
+function forces! end 
+
+function virial end 
+
+function energy_forces_virial end 
+
+function energy_forces_virial! end 
+
+
+
+
+
+


cortner:
@tjjarvinen -- can you please add a draft documentation page that explains the interface.

tjjarvinen:
I would actually add here that maybe the default should also include:

```julia
function forces!(f, data, calc; kwargs...)
    f .= forces(data, cal; kwargs...)
    return f
end

function energy_forces_virial(data, calc; kwargs...)
    e = energydata, calc; kwargs...)
    f = forces(data, calc; kwargs...)
    v = forces(data, calc; kwargs...)
    return (; :energy=>e, :forces=>f, :virial=>v)
end
```
etc. 

Also, I would say that `energy_forces` function should be part of the interface too.

cortner:
Yes, something like that should be added, I forgot to add this to the list at the top.

The reason I didn't yet add it to the code, is I'm not sure which should the default be? Don't we want the codes to implement `forces!` and then provide a wrapper 
```julia
forces(sys, calc) = forces!(alloc(...), sys, calc)
```
But then what about `forces_virial!`? There are so many variants.

cortner:
regarding `energy_forces` I'm not sure, couldn't one simply do 
```
E, F, _ = energy_forces_virial
```
but I dont have a strong view on this.

tjjarvinen:
No. Not all methods can calculate virial and you want those to throw error. This is case e.g. in isolated molecules, where virial is meaningless.

cortner:
One can equally argue that the virial is simply zero in that case since the energy doesn't change when you change cell shape.

tjjarvinen:
If you do constant volume simulation, you don't want to spend time to calculate virials. So, calling a method that calculates virials is a waste.

For me `energy_forces` and `energy_forces_virial` are the ones we should have, others are special cases.

cortner:
ok, let's assume this. But why `energy_forces` and not `energy_forces!`?  For electronic structure or large MLIPs models it doesn't matter but e.g. for a fast empirical potential the allocation might matter.

tjjarvinen:
I meant for the combinations only `energy_forces` and `energy_forces_virial` and the non-allocating ones for them.

 

> It has essentially zero extra cost if you already compute the forces.

For our ACE code it is not essentially free. It takes a considerable amount of time to gather the final value.

cortner:
I don’t think you are right - let’s benchmark this and stop arguing here.

cortner:
I think this debate had become unproductive. I’m closing the PR and will open a new one with a more succinct proposal.

Issue #87, opened 2023-10-02
Adding support for bonds and trajectories


ejmeitz:
Just wanted to start a discussion about potential ways to store bond information in the current AtomsBase object as well as how trajectories could be standardized.

- Bonds could just be added as an optional keyword with format up to the user or the format could be dictated. Either way there should be the option to include this data I think.
- Trajectories I'm a little less certain how to include as they are dynamic and AtomsBase fundamentally is not. That said a lot of analysis functions act on atomic trajectories not systems and it could be nice to unify trajectory formats as well.

cortner:
maybe those are two different discussions? Start a separate thread for the trajectories?  (btw, I only ever needed to treat trajectories as a list e.g. Vector or structures.)

rkurchin:
Also, a neighbor list is a bit different from bonds, since one has to do with spatial proximity and the other with specific chemical interactions (and, in the molecular simulation context, often geometric constraints).

ejmeitz:
Yes I'm more after adding function to the interface to standardize bonds in AtomsBase.

cortner:
To my mind this sort of interface necessarily involves neighbourlists. I thought it is the responsibility of the neighbourlist to provide lists of bonds - ideally lazily since anything beyond pair bonds would scale very badly. But if that's not how people commonly think about it, that's fine of course.

The op talks about "potential ways to store bond information" hence I asked about caching. If one wants to repeatedly access the list of bonds and if that list is provided via a neighbourlist, then I think this is related.

jgreener64:
In molecular mechanics terms the bonds are defined during setup and fixed throughout the whole simulation. The question is whether we should represent that information (a graph, effectively) in AtomsBase.

The neighbour list calculates close pairs from all possible atom pairs and is used to calculate non-bonded interactions (Lennard-Jones and Coulomb). There is the complexity that pre-defined bonds are important in this context, since atoms linked by one or two bonds typically have their interactions excluded and atoms linked by three bonds typically have their interactions down-weighted (huge hacks).

cortner:
> In molecular mechanics terms the bonds are defined during setup and fixed throughout the whole simulation.

Thank you for clarifying this for me. If that's the concrete context, I still think this is something to be calculated in a separate object and then stored somewhere. But why in a structure and not in the calculator? I can see arguments for both. Anyhow, any  object can already be stored? The calculator then needs to know how to find it. 

[...]

PR #90, opened 2023-11-18
add bonds to AbstractSystem

ejmeitz:
Added `bonds` function to `AbstractSystem` and an example of a `BondedSystem`. The example is basically just a `FastSystem`. I have not added tests to this yet.

This is based on discussion from #87.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -3,7 +3,7 @@ import PeriodicTable
 
 export AbstractSystem
 export BoundaryCondition, DirichletZero, Periodic, infinite_box, isinfinite
-export bounding_box, boundary_conditions, periodicity, n_dimensions, species_type
+export bounding_box, boundary_conditions, periodicity, n_dimensions, species_type, bonds
 export position, velocity, element, atomic_mass, atomic_number, atomic_symbol
 export atomkeys, hasatomkey
 
@@ -50,6 +50,23 @@ function boundary_conditions end
 Return the type used to represent a species or atom.
 """
 function species_type end
+
+"""
+The possible bond orders that can be used in the `bonds` function
+"""
+@enum BondOrder begin
+    single
+    double
+    triple
+end 
+
+"""
+    bonds(::AbstractSystem)
+
+Returns a Vector{Tuple{Integer, Integer, BondOrder}} where each entry stores the unique indices
+ i and j of the atoms participating in the bond and the order of the bond (e.g. [(13,21,5)])
+"""
+function bonds end
 
 """Return vector indicating whether the system is periodic along a dimension."""
 periodicity(sys::AbstractSystem) = [isa(bc, Periodic) for bc in boundary_conditions(sys)]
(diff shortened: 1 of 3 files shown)


src/interface.jl
```diff
@@ -51,6 +53,23 @@ Return the type used to represent a species or atom.
...
+    triple
+end 
+
+"""
+    bonds(::AbstractSystem)
+
+Returns a Vector{Tuple{Integer, Integer, BondOrder}} where each entry stores the unique indices
+ i and j of the atoms participating in the bond and the order of the bond (e.g. [(13,21,5)])
+"""
+function bonds end
```
rkurchin, line 71: Can you add a default implementation that just returns an empty tuple? e.g. instead of this line, something like:

```julia
bonds(sys::AbstractSystem) = [()]
```
(I'm not sure if there's a type stability issue here with that return type not having the type parameters of the Tuple, though)

src/interface.jl
```diff
@@ -51,6 +53,23 @@ Return the type used to represent a species or atom.
...
+    single
+    double
+    triple
+end 
+
+"""
+    bonds(::AbstractSystem)
+
+Returns a Vector{Tuple{Integer, Integer, BondOrder}} where each entry stores the unique indices
+ i and j of the atoms participating in the bond and the order of the bond (e.g. [(13,21,5)])
```
rkurchin, line 69: I don't think the example at the end of this line is actually in line with what you want, since the third element should be a `BondOrder` enum...

src/interface.jl
```diff
@@ -51,6 +53,23 @@ Return the type used to represent a species or atom.
...
+@enum BondOrder begin
+    single
+    double
+    triple
+end 
+
+"""
+    bonds(::AbstractSystem)
+
+Returns a Vector{Tuple{Integer, Integer, BondOrder}} where each entry stores the unique indices
```
rkurchin, line 68: Strictly, `Integer` should probably be `typeof(firstindex(sys))` (in practice, I think all implementations do use integer indices right now, but this is more general)

src/interface.jl
```diff
@@ -51,6 +53,23 @@ Return the type used to represent a species or atom.
 """
 function species_type end
 
+"""
+The possible bond orders that can be used in the `bonds` function
+"""
+@enum BondOrder begin
+    single
+    double
+    triple
```
rkurchin, line 62: should probably do this as
```julia
@enum BondOrder begin
    single = 1
    double = 2
    triple = 3
end
```

because it seems that if you don't do that, it actually assigns them numbers in a zero-indexed fashion (super weird), so the following can happen, which would be very confusing:

```julia
julia> BondOrder(2)
triple::BondOrder = 2
```

Separately, there might be a longer-term discussion about whether we should break out bonding types to be part of the interface since I could imagine people wanting more flexibility here, but my opinion is we start with this and file an issue for future conversations about this.

rkurchin (changes requested):
Left some comments, focusing on the interface itself for now. Once we have consensus on that, we should also make sure to add some tests for the `BondedSystem` type.

jgreener64:
My suggestion would be to not have the bond order as a required part of the interface. Without this it becomes a more general interface for storing atom pairs.

Instead we could have properties of bonds like we have properties of atoms, and some way to ask for the properties when calling `bonds`. This would also allow the storing of other properties if desired.

Issue #99, opened 2024-05-15
Introduce Cell types


cortner:
This is a proposal based on the zoom call on 15 / 5 / 2024, that replaces #97 . 

(1) remove `bounding_box` and `boundary_conditions` from the system and replace it with a cell object that encodes the same information. 
(2) Implement a type `Cell` or `ParaCell` that contains fields `cell_vectors` and `pbc` (true / false)
(3) At the same time I would like to replace the `AbstractVector` with `Vector` in order to remove that type instability as well.

It would look something like this: 
```julia
struct FlexibleSystem{D, TPART, TCELL} <: AbstractSystem{D}
    particles::Vector{TPART}
    cell::TCELL
    data::Dict{Symbol, Any}
end

struct ParaCell{D, T}
    cell_vectors::NTuple{D, SVector{3, T}}
    pbc::NTuple{D, Bool}
end
```

I think this will do two things: 
- make the `FlexibleSystem` type stable
- increase flexibility

I'll start on the PR if I get tentative agreement from a few people.

rkurchin:
I don't really have particularly strong feelings about how exactly we do things in the `FastSystem` and `FlexibleSystem` example implementations, but I presume this has a corresponding proposal for how the interface functions change?

I'm presuming based on our conversation earlier:
* get rid of [these lines](https://github.com/JuliaMolSim/AtomsBase.jl/blob/ba4f510898a48f5429e6f955cb3ce2b70bbfefe8/src/interface.jl#L13C1-L15C57) entirely
* change [this](https://github.com/JuliaMolSim/AtomsBase.jl/blob/ba4f510898a48f5429e6f955cb3ce2b70bbfefe8/src/interface.jl#L43) so that it returns a vector of `Bool`s with a length equal to `D`?

cortner:
I would try to stay as close as possible to the current interface. I don't know exactly what changes are needed until I work on this.

mfherbst:
I'm ok with this, provided that we have enough sugar that I don't have to go down to the cell datastructure when I just want to get the cell vectors out of a system. Because that's one thing I really dislike about the packages that abstract away the vectors in a cell structure is that this makes working with them quite inconvenient at times.

cortner:
> I would try to stay as close as possible to the current interface.

jgreener64:
I am okay with this, I don't have strong feelings about how `FastSystem` and `FlexibleSystem` are implemented. As Rachel says it is more about the interface functions.

Issue #97, opened 2024-05-13
Boundary Conditions 


cortner:
### PROPOSAL

Replace 
```
boundary_conditions::SVector{D, BoundaryCondition}
```
with one of 
```
# option 1  (strong preference for this)
pbc::NTuple{D, Bool}

# option 2
struct PBC{D}
   vals::NTuple{D, Bool}
end 

boundary_condition::PBC{D}
```

and to clarify in the documentation the difference between boundary conditions and constraints. 

This would be a breaking change and affect downstream packages.

### ORIGINAL POST: 

The boundary conditions currently cannot take two values without type instability: 
```
boundary_conditions::SVector{D, BoundaryCondition}
```
If I want PBC in x, y but open in z direction? A possible solution is to allow `Periodic(true)` and `Periodic(false)`. But e.g. `Periodic(), Open()` will not work. As I mentioned on Zulip, I think for particle systems there is only one type of boundary condition: periodic {true, false}. Everything else is to be implemented as a contraint. I propose to remove `boundary_conditions` entirely and replace it by a `pbc::NTuple{D, Bool}`.

`ZeroDirichlet` or anythig like that should be implemented as a contraint, not a boundary conditions. (this is different from continuum mechanics)

There are some more complicated cases that I'm not 100% certain how to address. E.g. reflections. (quadrupole method) but I suspect those are easiest to implement by just storing a larger system + the information that this systems has a certain symmetry

rkurchin:
This sounds fine to me, but I would want to get thoughts from folks who do other kinds of simulations in case there's something I'm not thinking of.

Conceptually, I think there may well be potential users who are not crystal clear about the distinction between BC's and constraints (this claim is based on some colloquial/sloppy language regarding these ideas that I certainly hear with some frequency), so whatever we settle on, we should add some clear language in both comments and documentation.

cortner:
A comment from @mfherbst  was that ZeroDirichlet is used for electronic structure. I agree this makes sense there, but I pointed out that that's a boundary condition for the electronic fields and not part of the particle system. My strong preference is to keep the particle system just a particle system and to not incorporate other properties into it.

cortner:
> there may well be potential users who are not crystal clear about the distinction between BC's and constraints

This is certainly true. I use the term boundary condition myself in my papers for things that I implement as constraints. We could avoid the term "boundary condition" entirely and call it something else.

cortner:
[...]

Issue #49, opened 2022-08-04
Chemical Element


cortner:
At the moment, the `Atom` type is defined as follows: 
```julia
struct Atom{D, L<:Unitful.Length, V<:Unitful.Velocity, M<:Unitful.Mass}
    position::SVector{D, L}
    velocity::SVector{D, V}
    atomic_symbol::Symbol
    atomic_number::Int
    atomic_mass::M
    data::Dict{Symbol, Any}  # Store arbitrary data about the atom.
end
```
I notice two things: why do we need atomic symbol AND atomic number? Doesn't it define the same thing? Secondly, `Symbol` is not an `isbbits` type (which really tripped me up a little while ago). 

I therefore propose to implement a (could be named something else of course) 
```julia 
struct ChemicalSpecies
    Z::Int 
end 
```
which is now `isbits`, and could be displayed as the symbol for readability.

rkurchin:
I agree.

The original reason for separating out some of these fields was to allow for a single particle to not necessarily represent an atom, but instead be more general to be e.g. a molecule or some other coarse-grained object. I think that's still possible, and one would then just not make use of the `Atom` type at all (and, correspondingly, `species_type` on the system would return something different.

We may still want to hold onto the `data` field, though, since otherwise `ChemicalSpecies` would essentially just be a rebrand of `Element` from PeriodicTable.jl.

cortner:
Maybe this now goes back to the #43 discussion. Personally, I would really prefer [something like this](https://github.com/ACEsuit/ACE.jl/blob/main/src/states.jl)

cortner:
> We may still want to hold onto the data field, though, since otherwise ChemicalSpecies would essentially just be a rebrand of Element from PeriodicTable.jl

I would want the species to be ultra-light-weight for speed. So yes it represents the same thing but it is an entirely different implementation. In fact it could be used to index into the list of elements.

cortner:
btw, the point about wrapping an integer is simply to distinguish the atomic number from an integer. There is no arithmetic on atomic numbers, one may want to dispatch on indexing, etc. I did this in JuLIP because my codes started getting confused when an integer was an index and when was it an atomic number.

rkurchin:
[...]

PR #101, opened 2024-05-18
WIP: Towards Type Stability

cortner:
This PR is to address a number of itches I have with the current implementation of the interface and its implementation. ~~So far the goal is to keep it entirely backward compatible. I think this may be possible.~~  I believe I am still very close to backward compatible in the interface, though no longer in the internals. 

The plan is to address the following issues: 
- [x] #99 
- [x] #97 
- [x] #49 
- [x] #5 
- [x] some discussions elsewhere about having a type-stable and isbits atoms type.
- [x] an ad-hoc decision to make `FlexibleSystem` mutable since it is supposed to be flexible? This will be good for `AtomsBuilder`
- [x] #102 

This is now ready for comments.

### Bugs and TODOs 

- [ ] make all tests pass
- [ ] write tests to show equivalent when using `FastAtom` 
- [ ] write tests that show performance when using `FastAtom` 
- [ ] cannot reinterpret `ChemicalElement` as `UInt8`
- [ ] I would like to delete or deprecate `DirichletZero`
- [ ] I would like to replace `AtomView` with `FastAtom` for `fast_system[i]`. The latter is "isbits" and hence performant. Also much much easier to use and implement, basically working like an `SVector`.

### DIFF
--- a/src/AtomsBase.jl
+++ b/src/AtomsBase.jl
@@ -5,9 +5,11 @@ using StaticArrays
 using Requires
 
 include("interface.jl")
+include("chem.jl")
 include("properties.jl")
 include("visualize_ascii.jl")
 include("show.jl")
+include("cell.jl")
 include("flexible_system.jl")
 include("atomview.jl")
 include("atom.jl")
--- a/src/atom.jl
+++ b/src/atom.jl
@@ -29,14 +29,17 @@ atomic_number(atom::Atom) = atomic_number(atom.chemical_element)
 element(atom::Atom)       = element(atom.chemical_element)
 n_dimensions(::Atom{D}) where {D} = D
 
-Base.getindex(at::Atom, x::Symbol) = hasfield(Atom, x) ? getfield(at, x) : getindex(at.data, x)
-Base.haskey(at::Atom,   x::Symbol) = hasfield(Atom, x) || haskey(at.data, x)
-function Base.get(at::Atom, x::Symbol, default)
-    hasfield(Atom, x) ? getfield(at, x) : get(at.data, x, default)
-end
+Base.getindex(at::Atom, x::Symbol) = 
+        hasfield(Atom, x) ? getfield(at, x) : getindex(at.data, x)
+Base.haskey(at::Atom, x::Symbol) = 
+        hasfield(Atom, x) || haskey(at.data, x)
+Base.get(at::Atom, x::Symbol, default) = 
+        haskey(at, x) ? getindex(at, x) : get(at.data, x, default)
+
 function Base.keys(at::Atom)
     (:position, :velocity, :chemical_element, :atomic_mass, keys(at.data)...)
 end
+
 Base.pairs(at::Atom) = (k => at[k] for k in keys(at))
 
 """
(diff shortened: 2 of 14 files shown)


cortner:
Recording changes
- Introduce `PCell` type to encode a computational cell with pbc flags
- Introduce `SystemWithCell{D, TCELL}` to help with dispatch 
- Adapt `FlexibleSystem` and `FastSystem` to accept `PCell` instead of `bounding_box` and `boundary_conditions`
- Introduce `FastAtom` to be compatible with `FastSystem`
- Made `FlexibleSystem` mutable.

cortner:
Question for the community: If I define a system that has open bc in some or all directions (i.e. is infinite in those directions) then the cell-vector pointing in that direction technically doesn't really specify a bounding box anymore. Should the call to bounding box in that case compute the actual extent of the system as specified by the positions of the particles? 

My intuition is that this is something that the neighbourlist should do and not the system. I would consider returning the cell vector or `NaN` or `Inf` in those cases. But I do not have a strong view yet.

cortner:
@mfherbst  Do you mind explaining this test to me: 
```julia
@testset "Chemical formula with system" begin
    lattice     = [12u"bohr" * rand(3) for _ in 1:3]
    atoms       = [Atom(6, randn(3)u"Å"; atomic_symbol=:C1),
                   Atom(6, randn(3)u"Å"; atomic_symbol=:C2),
                   Atom(1, randn(3)u"Å"; atomic_symbol=:D),
                   Atom(1, randn(3)u"Å"; atomic_symbol=:D),
                   Atom(1, randn(3)u"Å"; atomic_symbol=:D),
                  ]
    system      = periodic_system(atoms, lattice)
    @test atomic_symbol(system) == [:C1, :C2, :D, :D, :D]
    @test element_symbol(system) == [:C, :C, :H, :H, :H]
    @test chemical_formula(system) == "C₂H₃"
end
```

What is the point of `C1, C2`?

cortner:
My last question for now: Are the printing tests really needed? Why would we expect the ASCII output to remain unchanged when we make changes to the code?

tjjarvinen:
I would add isotopes to `ChemicalElement`.

Other issue that is related to isotopes is that in some cases you want to flag atoms to be different. E.g. you could say that these atoms are can be transformed to each other with symmetry operations and you then flag them with somehow. Another option is to flag atoms based on different chemical environment.

One way to add this would add extra field to chemical element. Albeit it could be moved to somewhere else too.

```julia
stuct ChemicalElement
   Z::UInit16
   N::UInt16 
   extra::UInt32
end
```

Note, I prefer using number of nucleons over number of neutrons, as then number 0 could be interpret, as field ignored. While with number of neutrons this is not possible.

The total bit size of the structure should be 64, 32, 16 or 128. For 64-bit system the structure is machine level will always be multiple of 64, so that size would be good target.

cortner:
I don't want to make `ChemicalElement` too rich - that goes against the idea of a simple prototype implementation. But we can absolutely have an extended chemical element type.

tjjarvinen:
Cache line is 64 bytes. So, 3\*Float64 for positions, mass is FLoat64, velosity is another 3\*Float64. If on top of this chemical element is 64-bits then it totals 64 bytes. Meaning that it would very efficient to move in memory. I would definitely go for this.

We could simply implement

```julia
 stuct ChemicalElement
   Z::UInt64
end

stuct ChemicalElementExt
   Z::UInt16
   N::UInt16 
   extra::UInt32
end
```

With these you can reinterpret the simple `ChemicalElement` to `ChemicalElementExt` that has option to have the extra features. Thus we have the default option to go for the simple `ChemicalElement`. While still have the option to define isotopes and additional features.

cortner:
In view of discussions in other channels I think this PR will likely be closed unmerged and some of the points raised here brought back after the interface update.

cortner:
closing since #107  is now at about the same stage as this old PR

Issue #113, opened 2024-09-03
Some ChemicalSpecies constructors create nneut=0


cortner:
```juliarepl
julia> ChemicalSpecies(:Si)
Si

julia> ChemicalSpecies(14)
Si14
```

cortner:
fixed by #114

Issue #112, opened 2024-09-02
atomic_number(::Int16) is undefined


cortner:
(this came up in tests for ExtXYZ)

Presumably it should always be `atomic_number(z::Integer) = z`

cortner:
fixed by #114

Issue #109, opened 2024-08-28
AbstractSystem not exported any more in v0.4


mfherbst:
I was a bit surprised to notice that `AbstractSystem` is no longer exported. I think this is a little unexpected (especially since `FastSystem` and `FlexibleSystem` are still exported), but I can't remember what we agreed on.

@cortner On purpose or a bug ?

cortner:
On purpose. I probably forgot to flag this. Sorry about that. What would be the reason to export it?

mfherbst:
For dispatches I use it in a couple of places.

mfherbst:
e.g. in DFTK to distinguish when an AbstractSystem or some DFTK-internal thing is used.

rkurchin:
I agree, I think it makes sense to export it for dispatch purposes, could imagine this being useful also for things like plotting recipes.

cortner:
Why is exporting needed for dispatch? Simply `using AtomsBase: AbstractSystem` in  any library that uses `AbstractSystem`. 

But as I'm thinking about it, then one use-case I see are update-constructors. Are there any left that are called via `AbstractSystem`? But tbh I would also prefer to rename those to something else. It doesn't feel natural.

mfherbst:
@cortner I see thanks for sharing your thoughts. I agree with your philosophy.

But I would argue that AbstractSystem is user-oriented. Update constructor is a potential case, but I also think about another case of  a user who is faced with a non AB datastructure, that they want to use in the AB ecosystem. For that standardising on a convert(AbstractSystem, datastructure) pattern is quite useful as the implementor of this method can still choose *which* AB backend to convert to ... and its not sth the user needs to know.

I have used this pattern in ASEconvert and in the Chemfiles so far, where the way to use python datastructures in julia (pyconvert(AbstractSystem, ase_atoms) also needs this implemented.

cortner:
I can see some arguments for update constructors but don't like using `AbstractSystem` as the symbol. I do not like the arbitrary selection of a default implementation. 

We can export it for now and discuss this again at the molssi workshop. @mfherbst if you make a quick PR that just adds an export line then just go ahead and merge and register without review.

cortner:
fixed by #114

PR #114, opened 2024-09-03
Several Minor Bugfixes

cortner:
- #113
- #112
- #109

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AtomsBase"
 uuid = "a963bdd2-2df7-4f54-a1ee-49d51e6be12a"
 authors = ["JuliaMolSim community"]
-version = "0.4"
+version = "0.4.1"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/AtomsBase.jl
+++ b/src/AtomsBase.jl
@@ -5,7 +5,7 @@ using UnitfulAtomic
 using StaticArrays
 using Requires
 
-export Atom, FlexibleSystem, FastSystem
+export Atom, FlexibleSystem, FastSystem, AbstractSystem 
 
 # Main Interface specification and inline docs 
 include("interface.jl")
--- a/src/utils/chemspecies.jl
+++ b/src/utils/chemspecies.jl
@@ -51,7 +51,10 @@ end
 
 Base.Broadcast.broadcastable(s::ChemicalSpecies) = Ref(s)
 
-ChemicalSpecies(z::Integer) = ChemicalSpecies(z, 0, 0)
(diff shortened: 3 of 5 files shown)


cortner:
@mfherbst @rkurchin -- can one of you take a brief look and confirm this is ok to merge?

cortner:
```
julia> O = AtomsBase.element(8)
julia> O.
appearance     atomic_mass    boil           category
color          cpk_hex        density        discovered_by
el_config      melt           molar_heat     name
named_by       number         period         phase
shells         source         spectral_img   summary
symbol         xpos           ypos
```
Not sure where in here I can extract the number of neutrons?

cortner:
@mfherbst @rkurchin -- sorry to ping you again, this PR is just a few lines. I'd be very grateful to get this approved since it blocks ExtXYZ from being updated to AB_0.4

rkurchin:
The specific answer to your chemistry question is that the reported mass is a weighted average over the different naturally-occurring isotopes of an element. For example, here's the table of this information from [Wikipedia for hydrogen](https://en.wikipedia.org/wiki/Isotopes_of_hydrogen):
image
(the reason that the standard atomic weight below the table is not exactly equal to .999855+.000145*2 is, I think, to do with the fact that for historical reasons, these are nondimensionalized against carbon, which is also not _exactly_ 12 because obviously C-14 etc. exist...I don't endorse this system, but it's what we're stuck with)

In any case, I think the rounding operation you add here is fine as a default behavior – obviously, if people care about isotopic detail, they'll probably implement something themselves that explicitly allows that DoF to be set and perhaps draws from a different database that has that information.

Alternatively, if you want it, it looks like @Gregstrq [has actually made basically this thing](https://github.com/Gregstrq/IsotopeTable.jl) that supposedly interoperates with PeriodicTable, though it hasn't been updated in two years so not sure if anything in there is stale...

cortner:
Any other questions or concerns about the PR?

cortner:
thank you.

cortner:
> But we need to do a proper fix later on. Maybe at the MolSSI meeting. 

I definitely agree with this.

Gregstrq:
> Alternatively, if you want it, it looks like @Gregstrq [has actually made basically this thing](https://github.com/Gregstrq/IsotopeTable.jl) that supposedly interoperates with PeriodicTable, though it hasn't been updated in two years so not sure if anything in there is stale...

Hi, everybody. I would say, that new isotopes do not get discovered every day. What I mean is: one would not expect to update the table of isotopes often, unless there are some mistakes in the way the data was compiled together. And if you do find any, please let me know.

cortner:
@Gregstrq  -- would be great to get your input in #115  please.  I think it is not at all clear yet how to manage this well.

PR #118, opened 2024-09-19
Fix #117

### LINKED ISSUE #117: `periodic_system` tutorial is broken
Hi, I was trying out AtomsBase.jl v0.4 and found that some of the docstrings are out of date.
For example, in `periodic_system` the docstring recommends the following
```jl
using Unitful
using UnitfulAtomic
using AtomsBase
bounding_box = 10.26 / 2 * [[0, 0, 1], [1, 0, 1], [1, 1, 0]]u"bohr"
silicon = periodic_system([:Si =>  ones(3)/8,
                                  :Si => -ones(3)/8],
                                 bounding_box, fractional=true)
```
First, `bounding_box` is now exported, so perhaps rename the variable to `bbox`? Second, the call to periodic system errors with
```
ERROR: MethodError: no method matching FlexibleSystem(::Vector{Atom{3, Quantity{…}, Quantity{…}, Quantity{…}}}, ::Vector{Vector{Quantity{…}}}, ::Vector{Bool})

Closest candidates are:
  FlexibleSystem(::AbstractVector{S}, ::Tuple{Vararg{var"#s15", D}} where var"#s15"<:AbstractVector{L}, ::Union{Bool, Tuple{Vararg{Bool, D}}, AbstractVector{<:Bool}}; kwargs...) where {L<:(Union{Quantity{T, 𝐋, U}, Level{L, S, Quantity{T, 𝐋, U}} where {L, S}} where {T, U}), S, D}
   @ AtomsBase ~/.julia/packages/AtomsBase/4AWmO/src/implementation/flexible_system.jl:51
  FlexibleSystem(::AbstractVector, ::Any; kwargs...)
   @ AtomsBase ~/.julia/packages/AtomsBase/4AWmO/src/implementation/flexible_system.jl:73
  FlexibleSystem(::Any; bounding_box, periodicity, kwargs...)
   @ AtomsBase ~/.julia/packages/AtomsBase/4AWmO/src/implementation/flexible_system.jl:69

Stacktrace:
 [1] atomic_system(atoms::Vector{Atom{…}}, box::Vector{Vector{…}}, bcs::Vector{Bool}; kwargs::@Kwargs{})
   @ AtomsBase ~/.julia/packages/AtomsBase/4AWmO/src/implementation/utils.jl:25
 [2] periodic_system(atoms::Vector{Pair{…}}, box::Vector{Vector{…}}; fractional::Bool, kwargs::@Kwargs{})
   @ AtomsBase ~/.julia/packages/AtomsBase/4AWmO/src/implementation/utils.jl:92
 [3] top-level scope
   @ REPL[17]:1
Some type information was truncated. Use `show(err)` to see complete types.
```
This seems to be fixed by instead defining
```
bounding_box = 10.26 / 2 .* ([0, 0, 1], [1, 0, 1], [1, 1, 0]) .* u"bohr"
```

I gather that there are several breaking API changes in v0.4. Is there a summary of them in some place?

cortner:
- fix #117 
- started implementing tests of docstrings 
- improve type signatures to allow auto-conversion of cell vectors and PBC

The last point is the main change: introduced `AUTOBOX` and `AUTOCELL` union types, added those to various function type signatures. These unions describe which input types for cell vectors / bounding box and pbcs can be converted into the required format via `_auto_cell_vectors` and `_auto_pbc` (cf. `cell.jl`). E.g. internally the cell vectors are always stured as `NTuple{D, SVector{D, T}}` but it can be more convenient to construct them as 
```julia
[[ a, 0, 0 ], [0, b, 0], [0, c, d] ] * u"nm"
```
and `_auto_cell_vectors` will then convert.

### DIFF
--- a/src/implementation/fast_system.jl
+++ b/src/implementation/fast_system.jl
@@ -16,8 +16,9 @@ struct FastSystem{D, TCELL, L <: Unitful.Length, M <: Unitful.Mass, S} <: Abstra
 end
 
 # Constructor to fetch the types
-function FastSystem(box::NTuple{D, <: AbstractVector}, pbc::NTuple{D, Bool}, 
-                    positions, species, masses) where {D}
+function FastSystem(box::AUTOBOX, 
+                    pbc::AUTOPBC, 
+                    positions, species, masses)
     cϵll = PeriodicCell(; cell_vectors = box, periodicity = pbc)
     FastSystem(cϵll, positions, species, masses)
 end 
@@ -39,18 +40,20 @@ function FastSystem(system::AbstractSystem)
 end
 
 # Convenience constructor where we don't have to preconstruct all the static stuff...
-function FastSystem(particles, box, pbc)
-    D = length(box)
-    if !all(length.(box) .== D)
+function FastSystem(particles, box::AUTOBOX, pbc::AUTOPBC)
+    box1 = _auto_cell_vectors(box)
+    pbc1 = _auto_pbc(pbc, box1)
+    D = length(box1)
+    if !all(length.(box1) .== D)
         throw(ArgumentError("Box must have D vectors of length D=$D."))
     end
-    if length(pbc) != D
+    if length(pbc1) != D
         throw(ArgumentError("Boundary conditions should be of length D=$D."))
     end
     if !all(n_dimensions.(particles) .== D)
         throw(ArgumentError("Particles must have positions of length D=$D."))
     end
-    FastSystem(box, pbc, position.(particles), species.(particles), mass.(particles))
+    FastSystem(box1, pbc1, position.(particles), species.(particles), mass.(particles))
 end
 
 Base.length(sys::FastSystem)         = length(sys.position)
--- a/src/implementation/flexible_system.jl
+++ b/src/implementation/flexible_system.jl
@@ -54,14 +54,6 @@ function FlexibleSystem(
         pbc::AUTOPBC{D};
         kwargs...
     ) where {S, D}
-    # if periodicity isa Bool 
-    #     periodicity = ntuple(_ -> periodicity, D)
-    # else 
-    #     periodicity = tuple(periodicity...)
-    # end
-    # if !all(length.(box) .== D)
-    #     throw(ArgumentError("Box must have D vectors of length D"))
-    # end
     cϵll = PeriodicCell(; cell_vectors = box, periodicity = pbc)
     FlexibleSystem{D, S, typeof(cϵll)}(particles, cϵll, Dict(kwargs...))
 end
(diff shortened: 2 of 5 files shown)


cortner:
@lxvm  --  let me know if this addresses your issue.

cortner:
@rkurchin @mfherbst @tjjarvinen  --  I need one of your reviews please to merge.

src/implementation/flexible_system.jl
```diff
@@ -49,20 +49,12 @@ Construct a flexible system, a versatile data structure for atomistic systems,
...
-    if !all(length.(box) .== D)
-        throw(ArgumentError("Box must have D vectors of length D"))
-    end
-    cϵll = PeriodicCell(; cell_vectors = box, periodicity = periodicity)
+        particles::AbstractVector{S},
+        box::AUTOBOX{D},
+        pbc::AUTOPBC{D};
+        kwargs...
+    ) where {S, D}
+    cϵll = PeriodicCell(; cell_vectors = box, periodicity = pbc)
```
mfherbst, line 57: Is this uncode character on purpose ... find that a bit weird.

src/implementation/fast_system.jl
```diff
@@ -39,18 +40,20 @@ function FastSystem(system::AbstractSystem)
 end
 
 # Convenience constructor where we don't have to preconstruct all the static stuff...
-function FastSystem(particles, box, pbc)
-    D = length(box)
-    if !all(length.(box) .== D)
+function FastSystem(particles, box::AUTOBOX, pbc::AUTOPBC)
+    box1 = _auto_cell_vectors(box)
```
mfherbst, line 44: Not sure about adding the 1 as a suffix ... but ok adding sth like `_` is not neccessarily better.

src/implementation/flexible_system.jl
mfherbst: Also this is a bit weird. I think we should use uniform names for arguments and keyword arguments. Maybe change the definition of `PeriodicCell` ?

src/utils/cells.jl
```diff
@@ -75,6 +75,16 @@ end
 # ---------------------------------------------
 #     Utilities 
 
+# allowed input types that convert automatically to the 
+# intended format for cell vectors,  NTuple{D, SVector{D, T}}
+const AUTOBOX = Union{NTuple{D, <: AbstractVector}, 
+                  AbstractVector{<: AbstractVector}} where {D}
```
mfherbst, line 81: Maybe inline a bit more to have the types in the Union line up ?

mfherbst (commented):
Nice idea !

src/implementation/flexible_system.jl
cortner: Deliberate. I don't want to overwrite a function name so this is the cost of not using getters with get_

src/implementation/fast_system.jl
cortner: Happy to change if there is a sensible standard. Until then I'll resolve this.

src/utils/cells.jl
cortner: done in next commit.

mfherbst (approved):

cortner:
thank you

PR #121, opened 2024-10-22
Extend AtomsBaseTesting and AtomsBase test routines

mfherbst:
- Reintroduce some previously disabled test
- Various smaller fixes to AtomsBase
- Fixes to AtomsBaseTesting

@cortner @rkurchin Could some of you take a quick look ?

### DIFF
--- a/src/utils/show.jl
+++ b/src/utils/show.jl
@@ -56,11 +56,11 @@ function show_system(io::IO, ::MIME"text/plain", system::AbstractSystem{D}) wher
     end
 
     # TODO - Not working at the moment
-    # ascii = visualize_ascii(system)
-    # if !isempty(ascii)
-    #     extra_line && println(io)
-    #     println(io, "   ", replace(ascii, "\n" => "\n   "))
-    # end
+    ascii = visualize_ascii(system)
+    if !isempty(ascii)
+        extra_line && println(io)
+        println(io, "   ", replace(ascii, "\n" => "\n   "))
+    end
 end
 
 Base.show(io::IO, system::AbstractSystem) = show_system(io, system)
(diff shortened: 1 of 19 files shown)


test/flexible_system.jl
```diff
@@ -0,0 +1,57 @@
+using AtomsBase
+using Unitful
+using UnitfulAtomic
+using Test
+
+@testset "Fast system" begin
```
cortner, line 6: Should this be "Flexible System" ?

src/utils/show.jl
```diff
@@ -18,7 +18,9 @@ function show_system(io::IO, system::AbstractSystem{D}) where {D}
     print(io, ")")
 end
 
-function show_system(io::IO, ::MIME"text/plain", system::AbstractSystem{D}) where {D}
+# TODO Put the show_ascii behind Preferences
+function show_system(io::IO, ::MIME"text/plain", system::AbstractSystem{D};
+                     show_ascii=true) where {D}
```
cortner, line 23: I vote again to make the default `false`. 87% of the time the ascii plot is not useful I think.

cortner (approved):

src/utils/show.jl
mfherbst: Happy to do that. If it's a preference I can activate it for myself. Other thoughts ?

We could also have a "denseness limit" e.g. only show if less than 10 atoms or so or if less than 10% of the image will be covered by strings ... but again, should be made configurable.

mfherbst: @rkurchin @jgreener64

mfherbst: I reverted that change here. Will open follow-up PR to start discussion in #122.

PR #122, opened 2024-10-23
Introduce preference for automatic ascii visualization of system

mfherbst:
Follow-up to #121 to discuss ascii visualisation preferences. Ping @cortner @rkurchin @jgreener64 to get some opinions.

I personally find it quite convenient to automatically get this insight when working in the REPL or a notebook, but this may just be me. I also found it helps students ... but again I introduced it, so I'm biased.

Options in my opinion are:
- Don't do it, people have to do it by calling `visualize_ascii` manually each time. Pros: Clear what happens, least complexity; Cons: It's an extra function to learn for something where Julia has builtin mechanics.
- Use some preference to configure it, e.g. Just enable/disable, some threshold in terms of number of atoms (done right now) or some threshold in terms of "density" of atoms on the canvas (to avoid cluttered ascii art, which is useless); Pros: Everyone can do as they please, if properly documented with helper functions should be fairly intuitive; Cons: We have to agree on defaults because most people will not change them.

In fact there may even be some builtin show preferences in Julia one can use for the second point ... I vaguely remember that, but I have not checked.

Other thoughts ?

### DIFF
--- a/src/utils/show.jl
+++ b/src/utils/show.jl
@@ -6,17 +6,18 @@ Configures the printing behaviour of `show_system`, which is invoked when a rich
 display of an `AbstractSystem` is requested. This is for example the case in a Julia REPL.
 The following options can be configured:
 
-- `max_species_list`: Maximal number of species in a system (`length(system)`) to trigger a
-   listing of every species along with its Cartesian positions. Default 10
-- `max_species_visualize_ascii`: Maximal number of species in a system to trigger a representation
-   in the form of an ascii cartoon using `visualize_ascii`. Default 0, i.e. diseabled.
+- `max_particles_list`: Maximal number of particles in a system until `show_system`
+  includes a listing of every particle. Default: 10
+- `max_particles_visualize_ascii`: Maximal number of particles in a system
+  until `show_system` includes a representation of the system in the form of
+  an ascii cartoon using `visualize_ascii`. Default 0, i.e. disabled.
 """
-function set_show_preferences!(; max_species_list=nothing, max_species_visualize_ascii=nothing)
-    if !isnothing(max_species_list)
-        @set_preferences!("max_species_list" => max_species_list)
+function set_show_preferences!(; max_particles_list=nothing, max_particles_visualize_ascii=nothing)
+    if !isnothing(max_particles_list)
+        @set_preferences!("max_particles_list" => max_particles_list)
     end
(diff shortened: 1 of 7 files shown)


rkurchin:
I personally haven't made much use of it, and have occasionally found it distracting, but can also imagine cases where it would be useful as you say. I would be fine with either option you describe.

cortner:
I'm ok with either option. If we go with a settable default then I will obviously vote to have a slim display as the default.

mfherbst:
Ok, I added the preferences. By default `visualize_ascii` will not be called.

mfherbst:
If one of you gives your ok, I guess we can merge.

src/utils/show.jl
```diff
@@ -1,4 +1,36 @@
 using Printf
+using Preferences
+
+"""
+Configures the printing behaviour of `show_system`, which is invoked when a rich `text/htmal`
```
rkurchin, line 5: typo, presumably should be `html`

src/utils/show.jl
```diff
@@ -1,4 +1,36 @@
...
+display of an `AbstractSystem` is requested. This is for example the case in a Julia REPL.
+The following options can be configured:
+
+- `max_species_list`: Maximal number of species in a system to trigger a listing of every species
```
rkurchin, line 9: potential ambiguity here whether max number of species would mean maximum number of atoms or e.g. maximum number of distinct elements

rkurchin (commented):
Generally LGTM, should for sure fix the one tiny typo, and if people want to discuss the minor potential ambiguity, we can, but I'm also okay just merging this as-is since it's not a huge deal

src/utils/show.jl
mfherbst: Better now ?

cortner: the documentation is fine, but I don't like the usage of `species` here, my instinctive interpretation of those keywords before reading the docs is that you decide how to print based on how many different elements there are in the system exactly as Rachel said. 

Since we are not restricting to atoms though, I suggest `max_particles_*`

mfherbst: Sounds good, I agree.

rkurchin:
Is this basically good to go? I don't think it should clash with #130 AFAICT, so presumably once coverage issues are addressed, it can be merged? I might have some time to add those couple of tests this afternoon if you don't already have them locally or something...

PR #126, opened 2024-10-25
Include some special casing for chemical species

mfherbst:
Adds a few more special cases to the handling of chemical species plus a bit of cleanup.

`:X` is fairly common in atomic structure files as a "dummy atom" with atomic number `0`. We can either force everyone to drop such atoms before using our stuff or include support here.

Here I am proposing to add this special case along with `:T` for tritium.

Note that I'm changing the Z -> data cache from an array to a Dict. Not sure if that has big performance implications, one could also handle this differently. @cortner probably knows if this may be an issue.

### DIFF
--- a/src/utils/chemspecies.jl
+++ b/src/utils/chemspecies.jl
@@ -1,8 +1,4 @@
-
-
-
-# --------------------------------------------- 
-#  Simple wrapper for chemical element type 
+# Simple wrapper for chemical element type 
 
 import PeriodicTable
 using Unitful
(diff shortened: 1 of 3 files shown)


cortner:
> Note that I'm changing the Z -> data cache from an array to a Dict. Not sure if that has big performance implications, one could also handle this differently. @cortner probably knows if this may be an issue.

I doubt this matters in practice.

cortner:
I don't mind about `:T` and `:X`, I have no strong views. I do suggest though that any atoms of type `:X` have no units...

tjjarvinen:
Yes, I have plan and can make PR over weekend. After that we can discuss how to proceed.

mfherbst:
Ok sounds good.

mfherbst:
What's the status with respect to the overhaul of `ChemicalSpecies`?

The inclusion of `0 / :Z` is blocking for me in the move to AtomsBase 0.4. If redesigning `ChemicalSpecies` will take some time (fine for me), could we get this minimal modification merged ?

tjjarvinen:
I was busy moving to back Europe the last week. The new version only needs masses added for isotopes and few other small things. I should be able to do it by tomorrow.

mfherbst:
Superseded by #130

PR #127, opened 2024-10-25
Rename bounding_box to cell_vectors consistently

mfherbst:
Renamed occurrences of cell_vectors to bounding_box for consistency.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AtomsBase"
 uuid = "a963bdd2-2df7-4f54-a1ee-49d51e6be12a"
 authors = ["JuliaMolSim community"]
-version = "0.4.2"
+version = "0.5.0"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -23,8 +23,8 @@ ChemicalSpecies
 ## System properties
 
 ```@docs
-bounding_box
-set_bounding_box!
+cell_vectors
+set_cell_vectors!
 periodicity
 set_periodicity!
 cell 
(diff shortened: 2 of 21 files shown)


cortner:
I disagree with this change. `cell_vectors` is more typical than `bounding_box` and in fact I would remove `bounding_box` in favour of `cell_vectors`.

mfherbst:
Fine with me as long as it's consistent in the code base.

@jgreener64 @rkurchin others thoughts ?

rkurchin:
It seems that the core point here is that we should be consistent about what we call this thing, which I definitely agree with. I personally do not have strong feelings about which particular name we go with.

jgreener64:
No opinions beyond consistency.

cortner:
not sure what to say here, I stand by what I said. google "bounding box" and "cell vectors" and see what you get. Bounding box is a concept from image processing, while cell vectors immediately brings up links to [Wikipedia Unit Cell](https://en.wikipedia.org/wiki/Unit_cell).

If you all disagree with me and decide you prefer to stick with the existing terminology, then not much I can do. 

If you want to consider changing the terminology, then the natural thing to do would be to deprecate the `bounding_box` function for the next minor version.

tjjarvinen:
I would prefer `cell_vectors` also. It is a much better name, as it does not need explanations. (plus what Christoph mentioned about googling)

mfherbst:
Same here. If no one expresses clear preference, I'd suggest we change everything to `cell_vectors`.

mfherbst:
Ok, this will now completely remove `bounding_box` in favour of `cell_vectors`.

How do we feel about the plural form here. I used it because of our previous discussion. But now I note that `cell_vectors` is potentially inconsistent with our other functions and keyword arguments, which are all singular.

I'm not too sure about this, but wanted to seek your opinions as well.

rashidrafeek:
I think the plural form `cell_vectors` is fine as it is a common terminology. 

Also just pointing out that this is technically breaking as `bounding_box` was not removed from the interface in the v0.4 update and is still [mentioned](https://juliamolsim.github.io/AtomsBase.jl/stable/interface/#Core-Interface) as part of the interface. It doesn't affect me personally, as I have not yet shifted to 0.4, so no issues if we decide to remove it. Would it be better to just deprecate `bounding_box` and remove it in 0.5?

cortner:
Technically we can use cell_vector(sys, :) but I agree this is strange in this context. But it is not inconsistent because cell_vectors is a system property and not a particle property.

I strongly support plural.

(and yes to the last post - this is breaking and will have a deprecation cycle.)

mfherbst:
Sure, I agree. Should get to it later today.

mfherbst:
Unless there are objections I'll merge this tomorrow.

Issue #115, opened 2024-09-03
Revisit ChemicalSpecies


cortner:
(cf discussion in #114)

The `ChemicalSpecies` implementation has some problems, partly due to my lack of understanding of chemistry. Basically, the issue is that we now require the number of neutrons to be specified. But the correct interpretation when this is not done is that one uses the average number of neutrons (taken over all isotopes). I believe there are two relatively straightforward solutions to this issue: 

(1) Allow `n_neutrons = -1` as this default with unspecified `n_neutrons`

(2) Introduce a new type e.g. `Element` that only specifies the atomic number and nothing else.

I would be grateful for others' thoughts. 

CC @rkurchin @mfherbst @jgreener64 @tjjarvinen

tjjarvinen:
I propose a small extension to the current design. It would be a non-breaking one.

If following can be made happen, then everything that is needed for chemistry would be implemented. 

```julia
ce1 = ChemicalElement(6)  # Carbon atom
ce2 = ChemicalElement(:C)  # equal to above
ce3 = ChemicalElement(6; n_nucleons=13)
ce4 = ChemicalElement(6; n_nucleons=12)
ce5 = ChemicalElement(6; atom_name=:Cah)
ce6 = ChemicalElement(6; n_nucleons=13, atom_name=:Cah)
ce7 = ChemicalElement(6; n_nucleons=12, atom_name=:Cah)

ChemicalElement(:Cah) # error

ce1 == ce2 # true
ce1 == ce3 # true
ce1 == ce4 # true
ce1 == ce5 # false
ce3 == ce4 # false
ce3 == ce5 # true
ce3 == ce6 # true
ce4 == ce5 # true
ce4 == ce6 # true
ce4 == ce7 # false
ce5 == ce6 # true
ce5 == ce7 # true
ce6 == ce7 # false

atomic_symbol( ce ) # return :C for all above cases 

atomic_name( ce1 ) # return :C
atomic_name( ce2 ) # return :C
atomic_name( ce3 ) # return :C
atomic_name( ce4 ) # return :C
atomic_name( ce5 ) # return :Cah
atomic_name( ce6 ) # return :Cah
atomic_name( ce7 ) # return :Cah
```

### Reasoning

[...]

PR #130, opened 2024-11-03
Chemspecies update

tjjarvinen:
This is update to ChemicalSpecies.

- Change `info` field to `name` field and make it represent 4 characters
- Add function `atom_name` to access the new name field if given, else return `atomic_symbol`
- Added masses for isotopes
- Added symbol X to valid input. Gives weightless particle with atomic number zero

The idea with `atom_name` is to represent PDB file format [defines](http://www.wwpdb.org/documentation/file-format-content/format33/v3.3.html)  atom name columns (13-16). While `element_symbol` represents element columns (77-78)

1. `element_symbol` always gives the symbol of the element
2. `atomic_symbol` adds isotope information if given
3. `atom_name` defaults to `atomic_symbol`, but overwrites it with `atom_name` if species has that information.

### This works now

```julia
julia> ChemicalSpecies(:C)
C

julia> ChemicalSpecies(:C12)
C12

julia> ChemicalSpecies(:C; atom_name=:MyC)
C

julia> ChemicalSpecies(:C13; atom_name=:MyC)
C13

julia> ChemicalSpecies(:X)
X
````

### Comparisons

The basic idea is that if no isotope or name information is given, then that ChemicalSpecies works as a general atom with that atomic number.

`atom_name` on the other hand is used to distinguish  atoms with same type of atoms. 

```julia
# true
ChemicalSpecies(:C13) == ChemicalSpecies(:C)

# false
ChemicalSpecies(:C12) == ChemicalSpecies(:C13)

# true
ChemicalSpecies(:C; atom_name=:MyC) == ChemicalSpecies(:C)

# true
ChemicalSpecies(:C12; atom_name=:MyC) == ChemicalSpecies(:C12)

# false
ChemicalSpecies(:C; atom_name=:MyC) == ChemicalSpecies(:C12)

# true
ChemicalSpecies(:C12; atom_name=:MyC) == ChemicalSpecies(:C)

# true
ChemicalSpecies(:C; atom_name=:MyC) == ChemicalSpecies(:C12; atom_name=:MyC)
```

So basically, you can create a potential using `ChemicalSpecies(:C)`. Users can then mark their systems with isotopes or atomic names, for some additional information and the potential that uses "the general carbon" works with them.

### masses

```julia
julia> mass(ChemicalSpecies(:C))
12.011 u

julia> mass(ChemicalSpecies(:C12))
12.0 u

julia> mass(ChemicalSpecies(:C13))
13.003354835340001 u

julia> mass(ChemicalSpecies(:C; atom_name=:MyC))
12.011 u

julia> mass(ChemicalSpecies(:C13; atom_name=:MyC))
13.003354835340001 u

julia> mass(ChemicalSpecies(:X))
0.0 u
```

@mfherbst @cortner 

edit. changed `atomic_name` to `atom_name`


rashidrafeek:
just a suggestion: `atomic_name` seems to imply the name is a property of the atom itself instead of an identifier (which seems to be the case here). Wouldn't just `atom_name` be better which also matches the PDB defintion?

tjjarvinen:
This is now up for review and merge.


> just a suggestion: `atomic_name` seems to imply the name is a property of the atom itself instead of an identifier (which seems to be the case here). Wouldn't just `atom_name` be better which also matches the PDB defintion?

I agree with this one and made the change.

This is of course up for a debate, so feel free to state your opinions.

cortner (approved):

cortner:
I read through and at least for the use-cases I can envision at the moment I am happy. For me this is ok to merge for now. 

There is one small item worth thinking about : I find it odd to have a large table of isotope data in a julia source file rather than a JSON or similar. But this is just a matter of preference and I don't feel very strongly about it.

tjjarvinen:
Fixed a bug where `ChemicalSpecies(:U238)`  and similar did not work correctly.


> There is one small item worth thinking about : I find it odd to have a large table of isotope data in a julia source file rather than a JSON or similar. But this is just a matter of preference and I don't feel very strongly about it.

The reasoning here was that having JSON or similar would add more dependencies. So, having it in julia source allows minimizing dependencies.

src/utils/chemspecies.jl
```diff
@@ -11,9 +11,15 @@ import Base: ==, convert, show, length
 
 export ChemicalSpecies
 
+# masses are saved in a different file to not clutter this file
+include("isotope_masses.jl")
+
 """
 Encodes a chemical species by wrapping an integer that represents the atomic 
-number, the number of protons, and additional unspecified information as a `UInt32`. 
+number, the number of protons, and additional name as max 4 characters.
```
Luthaf, line 19: Sorry for the drive-by comment, but why are you limiting the name to 4 characters? This is the limit for the obsolete PDB format, but for example the mmcif format that replaced it does not have any length limitation: https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v40.dic/Items/_chem_comp_atom.atom_id.html

tjjarvinen: In order to have a bits type you need to have fixed amount of bits. This leads to a limitation on character length, which for now is 4*8bit char. We could of course increase it later on (there are several ways to do this). But the main design goal was to have it as bits type, which e.g., allows moving atom types to GPU etc.

Luthaf: Fair enough!

rkurchin:
This looks good to me also – should we link this with #115 so it will close when this is merged?

mfherbst (commented):
Very nice, thanks for the extension.

src/utils/chemspecies.jl
```diff
@@ -51,86 +88,98 @@ end
...
+                tmp = parse(Int, str_symbol[end-1:end])
+                str_symbol = str_symbol[1:end-2]
+            end
+        else
+            tmp = parse(Int, str_symbol[end])
+            str_symbol = str_symbol[1:end-1]
+        end
+    end
+    asymbol = Symbol(str_symbol)
+    z = haskey(_sym2z, asymbol) ? _sym2z[asymbol] : 0
```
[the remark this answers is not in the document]
tjjarvinen: I added a check for symbols and now symbols that are not recognized throw an ArgumentError.

tjjarvinen:
Everything should now be ok with this PR.

mfherbst (approved):
Lgtm

mfherbst:
Unless there are objections I'll merge this tomorrow.

PR #136, opened 2025-01-22
Fix bug in element_symbol call

tjjarvinen:
Currently this will fail

```julia
using AtomsBase
using Unitful

hydrogen = isolated_system([
               :H => [0, 0, 0.]u"Å",
               :H => [0, 0, 1.]u"Å"
           ])
element_symbol(hydrogen, 1)  # error
```

This PR fixes it.

### DIFF
--- a/src/utils/chemspecies.jl
+++ b/src/utils/chemspecies.jl
@@ -249,7 +249,7 @@ this may be different than `atomic_symbol` for cases where `atomic_symbol`
 is chosen to be more specific (i.e. designate a special atom).
 """
 element_symbol(sys::AbstractSystem, index) = 
-        element_symbol.(sys[index])
+        element_symbol.(species(sys,index))
 
 element_symbol(species) = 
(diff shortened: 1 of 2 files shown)


src/utils/chemspecies.jl
```diff
@@ -249,7 +249,7 @@ this may be different than `atomic_symbol` for cases where `atomic_symbol`
 is chosen to be more specific (i.e. designate a special atom).
 """
 element_symbol(sys::AbstractSystem, index) = 
-        element_symbol.(sys[index])
+        element_symbol.(species(sys,index))
```
mfherbst, line 252: Space after `,` please

mfherbst (commented):
Thanks @tjjarvinen !

src/utils/chemspecies.jl
tjjarvinen: done

PR #140, opened 2025-09-20
fix #139

### LINKED ISSUE #139: ChemicalSpecies and Symbol == only works one way
The example speaks for itself:
```julia-repl
julia> ChemicalSpecies(:Si) == :Si
true

julia> :Si == ChemicalSpecies(:Si)
false
```

Looks like while `==(a::ChemicalSpecies, sym::Symbol)` was implemented, a corresponding `==(sym::Symbol, a::ChemicalSpecies)` is missing.

tjjarvinen:
The issue had a little more complicated that what it looked like. There was another bug that triggered when when trying direct fix.

So. this fixes the original issue

```julia
julia> ChemicalSpecies(:Si) == :Si
true

julia> :Si == ChemicalSpecies(:Si) # now fixed
true
```

and the other issue that was

```julia
julia> ChemicalSpecies(:H) == :not_atom
ArgumentError
```

and now is

```julia
julia> ChemicalSpecies(:H) == :not_atom
false
```

For the record the latter was triggered by

```julia
tmp = ChemicalSpecies(:C12; atom_name=:MyC)
@test atom_name(tmp) != atomic_symbol(tmp)
```

which raised an error for `ChemicalSpecies(:MyC)`.

Because this is a bigger fix this needs some reviews too.

### DIFF
--- a/src/utils/chemspecies.jl
+++ b/src/utils/chemspecies.jl
@@ -139,9 +139,24 @@ end
 
 ChemicalSpecies(sym::ChemicalSpecies) = sym
 
-==(a::ChemicalSpecies, sym::Symbol) = 
-        ((a == ChemicalSpecies(sym)) && (Symbol(a) == sym))
+function ==(a::ChemicalSpecies, sym::Symbol)
+    # We need to catch ArgumentError here as e.g.
+    # ChemicalSpecies(:H) == :not_atom
+    # does throw an error but should return false
+    local tmp
+    try 
+        tmp = ChemicalSpecies(sym)
+    catch e
+        if e isa ArgumentError
+            return false
+        else
+            rethrow()
+        end
+    end
+    return (a == tmp) && (Symbol(a) == sym)
+end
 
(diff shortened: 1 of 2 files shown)


src/utils/chemspecies.jl
```diff
@@ -139,9 +139,24 @@ end
...
 
-==(a::ChemicalSpecies, sym::Symbol) = 
-        ((a == ChemicalSpecies(sym)) && (Symbol(a) == sym))
+function ==(a::ChemicalSpecies, sym::Symbol)
+    # We need to catch ArgumentError here as e.g.
+    # ChemicalSpecies(:H) == :not_atom
+    # does throw an error but should return false
+    local tmp
+    try 
+        tmp = ChemicalSpecies(sym)
```
mfherbst, line 148: Can't you just also do the return in here. This would simplify the code a bit ... or else assign the result of the `try` block to an argument and use that below ?

tjjarvinen: This is now
```julia
tmp = try
        ChemicalSpecies(sym)
catch e
```

PR #148, opened 2026-07-01
Add isless, isequal, hash for ChemicalSpecies

cortner:
Implementation of isless, isequal, hash for ChemicalSpecies. Original need for me was sorting, but then it became apparent that these are three canonical related things that are important for: sort, Dict, Set and similar.

Simple addition. Non-breaking.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AtomsBase"
 uuid = "a963bdd2-2df7-4f54-a1ee-49d51e6be12a"
 authors = ["JuliaMolSim community"]
-version = "0.5.2"
+version = "0.5.3"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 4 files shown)


tjjarvinen:
This is a needed addition and a good implementation!

There is a one issue that this raises and that is the case of 

```julia
h = ChemicalSpecies(:H) 
d = ChemicalSpecies(:D) 

h == d  # true
isequal(h, d) # false

# meaning `==` is not same as `isequal`
```

This is completely valid in Julia, but can cause some issues, if you don't know it before hand. So, it might be a good to add a note to the documentation that this exist.

We could even discuss on whether we should have a dedicated name for a function that does the current version of `==` now, so that we could have `==` and `isequal` to be the same. But that would be a breaking release.

cortner:
Fair point. Since making == and isequal behave the same is breaking I would document this, merge and then file an issue for the breaking change that you suggest.

cortner:
Based on @tjjarvinen 's comments I assume this is ok and will merge tomorrow unless I hear from somebody blocking it.

